PR #39, opened 2015-06-30
Fixed issue where the first type was always recognized as the global …

LMSDev:
…type. Not entirely about the detection method, but should do it. For assemblies where <Module> and other stuff is stripped, I had to add this to get Confuser to resolve properly. Nice project btw. - LMS

### DIFF
--- a/src/DotNet/ModuleDef.cs
+++ b/src/DotNet/ModuleDef.cs
@@ -467,9 +467,20 @@ namespace dnlib.DotNet {
 		/// <summary>
 		/// Gets the global (aka. &lt;Module&gt;) type or <c>null</c> if there are no types
 		/// </summary>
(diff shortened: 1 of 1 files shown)


LMSDev:
Apparently GitHub didn't like the word < Module > after "where".

0xd4d:
The global type, aka `<Module>` type is always the first type and can be named anything.

LMSDev:
Ah okay. Without this "fix" dnlib refuses to resolve types (rather the first/global type) for me though, because one of my references is an assembly where the < Module >  information has been stripped. Would you consider that as bad luck or is there a way we can integrate this?

0xd4d:
Could you email me a file? de4dot@gmail.com

PR #42, opened 2015-07-07
Added support for SNK

picrap:
Signature is different from PFX, so I changed the signature check.
(This is because I want to use the signature feature from ConfuserEx)


picrap:
Sorry I found some strange behavior. I'm closing this request right now, hoping to reopen it soon.

picrap:
My mistake, this was not necessary.

PR #44, opened 2015-08-08
DynamicBodyReader did not update method parameters which lead to wron…

grimaz:
…g body generation in some cases. Like starg_1 on .Read() appeared as starg null.


0xd4d:
Thanks.

PR #64, opened 2016-05-30
Removed check for empty body inside WriteMethodBodies

chaplin89:
Hi,
can I ask you what's the rationale behind this check? I've run into troubles because of that.
There are obfuscators (like ConfuserEx with "anti-tamper" protection) that replace bodies of methods on the fly after decrypting them, so even if bodies may appear empty it's important to save all the information related to bodies, in order to let these application to run correctly.

Thanks,
Marco


chaplin89:
Sorry, my fault. That was not the real problem.

PR #110, opened 2017-04-01
Add support for .NET Core build tools

kiootic:
As title

### DIFF
--- a/src/dnlib.csproj
+++ b/src/dnlib.csproj
@@ -16,6 +16,14 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\dnlib.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(OS)' != 'Windows_NT'">
+    <PublicSign>true</PublicSign>
+    <!-- Workaround for https://github.com/dotnet/sdk/issues/335 -->
+    <FrameworkPathOverride Condition="Exists('/usr/lib/mono/2.0-api')">/usr/lib/mono/2.0-api</FrameworkPathOverride>
+    <FrameworkPathOverride Condition="Exists('/usr/local/lib/mono/2.0-api')">/usr/local/lib/mono/2.0-api</FrameworkPathOverride>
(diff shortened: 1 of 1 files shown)


src/dnlib.csproj
```diff
@@ -15,6 +15,8 @@
     <FileAlignment>512</FileAlignment>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\dnlib.snk</AssemblyOriginatorKeyFile>
+    <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
```
0xd4d: Why is `PublicSign` needed to sign it?

src/dnlib.csproj
```diff
@@ -15,6 +15,8 @@
...
+    <FrameworkPathOverride Condition="'$(OS)' != 'Windows_NT'">/usr/lib/mono/2.0-api/</FrameworkPathOverride>
```
0xd4d: This is a hard coded path. Is it not possible for the build tools to find this path themselves?

0xd4d: Have you tried setting an environment variable instead? See https://github.com/dotnet/netcorecli-fsc/wiki/.NET-Core-SDK-rc4#using-net-framework-as-targets-framework-the-osxunix-build-fails

src/dnlib.csproj
kiootic: If it is not used, compilation will fail with 'Assembly signing not supported'.

src/dnlib.csproj
kiootic: I found this: https://github.com/dotnet/core-setup/blob/master/src/Microsoft.Extensions.DependencyModel/Resolution/DotNetReferenceAssembliesPathResolver.cs
Would it be ok to use multiple FrameworkPathOverride with condition of Path Exist?

0xd4d: Can you confirm if this fix is for [this bug](https://github.com/dotnet/sdk/issues/335) ? If so you can add the hard coded paths (with Path.Exists checks) since it appears that MS does the same thing. But add a comment in the csproj file pointint to the issue.

kiootic: It shows some path that .NET Core installer will search for, so I think it is ok to hard-code these path for now. I have added a comment link to the issue.

0xd4d:
Thanks!

Issue #86, opened 2016-12-09
Are there any plans to port this to .NET Core?


NickStrupat:
I took a peek already and it looks like a few things will need to change: serialization, memory-mapping...

0xd4d:
It's something I probably will do some day, but not in the near future.

zwcloud:
Since .NET Standard 2.0 has been released, is it planned now?

0xd4d:
.NET Core 2.0 added support to load .NET Framework assemblies. Have you tried it yet?

zwcloud:
Yes. Actually it has already been suported in .NET Core 1.0 or even earlier version.
To be exact, they are called [.NET Standard class library](https://docs.microsoft.com/en-us/dotnet/standard/class-libraries). For example, a .NET Standard class library can be used by all the platforms that support .NET Standard.

0xd4d:
I'm talking about non-modified .NET Framework assemblies. This is new in .NET Core 2.0

zwcloud:
I just tried adding a .NET Framework 4.6.1 Class libarary to my .NET Core 2.0 App. It works.

But I should say, in theory, if the class library is built on .NET Framework 4.5+, it can be used with .NET Core projects. This is not new in .NET Core 2.0.

zwcloud:
BTW, here is an issue you may be interested in: https://github.com/dotnet/coreclr/issues/13563
I'm stuck when trying to use ModPhuserEx( successor of ConfuserEx). I'm sure it is a .NET Core 2.0 specific issue. _dnlib_ works well on .NET Core 1.1/.NET Standard 1.6 assemblies.

> Exception: dnlib.DotNet.AssemblyResolveException: Could not resolve assembly: netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51

zwcloud:
> dnlib can't do anything about missing files.

No, it is not a missing file. I can't tell what the problem is but it seems .NET Core makes some modification to assembly loading/referancing procedures, which may possibly become a **trap** when you trying to port dnlib to .NET Core/Standard. I just hope this information will help you in further developing.

see also:
https://stackoverflow.com/questions/45021965/could-not-load-file-or-assembly-netstandard-version-2-0-0-0-reference
https://developercommunity.visualstudio.com/content/problem/97436/fakes-with-net-standard-20.html

0xd4d:
[...]

PR #148, opened 2017-12-04
Patch for .NET Core 2.0

CaffeinatedCoder:
Possible solution for #86

Added new csproj file targeting .NET Core 2.0. Written in JetBrains Rider under MacOS, compiles without problems in Visual Studio 2017 and under MacOS in Rider. Ran .NET Portability Analyzer against the source code and compiled binary: 100% .NET Core 2.0 compatibility.


.gitignore
```diff
@@ -8,3 +8,8 @@
 /Release/
 /Examples/bin/
 *.tmp_proj
+src/.idea/.idea.src/riderModule.iml
```
0xd4d: `src/.idea/` is enough to exclude everything.

src/ExtensionAttribute.cs
```diff
@@ -2,8 +2,10 @@
 
 #pragma warning disable 1591	// XML doc warning
 
+#if !NETCOREAPP2_0
```
0xd4d: Why isn't this `NETSTANDARD2_0`?

.gitignore
CaffeinatedCoder: You are right :-)

src/ExtensionAttribute.cs
CaffeinatedCoder: This is just the name of the compile flag inside the csproj file, automatically given by Rider while creating an empty project file. The name corresponds to the referenced platform, which is basically *Microsoft.NETCore.App/2.0.0*.

src/dnlib.netcore20.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
```
0xd4d: It's a library and should target netstandard, not netcoreapp

.gitignore
```diff
@@ -8,3 +8,4 @@
 /Release/
 /Examples/bin/
 *.tmp_proj
+src/.idea/*
```
0xd4d: The asterisk isn't needed, see the other lines

src/dnlib.netcore20.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <Product>dnlib</Product>
+    <RootNamespace>dnlib</RootNamespace>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <AssemblyName>dnlib</AssemblyName>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <OutputPath>..\Debug\bin\</OutputPath>
```
0xd4d: Won't this overwrite the other non-netstandard2.0 file?

.gitignore
0xd4d: `.idea/` is probably all that's needed since it seems like that thing could be added to any dir

src/dnlib.netcore20.csproj
CaffeinatedCoder: No, I tested that as well. .NET Core binaries always end up in a separate "netcoreapp2.0" sub-folder under *../{Debug | Release}/bin*.

src/dnlib.netcore20.csproj
CaffeinatedCoder: netcoreapp2.0 is not meant for applications, it means this library targets explicitly .NET Core 2.0 as a target framework. Changing this to netstandard2.0 manually (IDEs don't provide this option) doesn't compile, cause there are types and methods missing from System.Reflection and System.Runtime.InteropServices in .NET Standard 2.0, which are to be found in .NET 4.x and .NET Core 2.0.
I can dive into this deeper, but no promise if this can be solved:

**dnlib.netcore20
  MethodTableToTypeConverter.cs(89, 27): [CS1061] 'TypeBuilder' does not contain a definition for 'CreateType' and no extension method 'CreateType' accepting a first argument of type 'TypeBuilder' could be found (are you missing a using directive or an assembly reference?)
  MethodTableToTypeConverter.cs(108, 27): [CS1061] 'TypeBuilder' does not contain a definition for 'CreateType' and no extension method 'CreateType' accepting a first argument of type 'TypeBuilder' could be found (are you missing a using directive or an assembly reference?)
  ModuleDefMD.cs(246, 26): [CS0117] 'Marshal' does not contain a definition for 'GetHINSTANCE'**

I would suppose to introduce a NuGet package and to provide different project files for different targets, when installing from NuGet you will then always get the right dll provided for your solution. As of now, pure netstandard2.0 is not possible unfortunately.

0xd4d: I don't know how useful a netcoreapp build would be since .NET Core 2.0 can already load .NET Framework assemblies, most likely also dnlib.dll. A netstandard build however is much more useful since it doesn't require .NET Core.

.NET Framework, .NET Core and Mono all support netstandard 2.0, but netstandard 2,0 is missing some APIs used by dnlib.

I'll have a look at this some time when I have more time and interest. It's better to just support netstandard 2.0 since it can be used by all three major .NET runtimes.

0xd4d:
Fixed by  #155

PR #155, opened 2018-02-20
Port to netstandard2.0

SlowLogicBoy:
For some reason I can't build solution using VS2017, so I had to use VS2015, that's why I didn't add `dnlib.netstandard.csproj` to solution and so it wouldn't break anything.

To build `dnlib.netstandard.csproj`:
```
cd src
dotnet build dnlib.netstandard.csproj
```
Did as less changes as I could.
dnlib.sln should build the same way as it did before.

I tried porting it to netstandard1.1;netstandard1.3;netstandard1.6 with no success, so I left it to netstandard2.0.

Used some code from #148 
**Issues:**
1. **Loading from `System.Reflection.Module` doesn't work on netcoreapp2.0 when ran on other than Windows platforms (Mono should work).**

TODO:
- [ ] Test on Mono.
- [ ] Test on netcoreapp2.0 on linux


.gitignore
```diff
@@ -1,11 +1,11 @@
 *~
-*/obj/
```
0xd4d: .gitignore doesn't need to be updated, it still hides all build files

src/DotNet/ModuleDefMD.cs
```diff
@@ -234,7 +234,20 @@ public sealed class ModuleDefMD : ModuleDefMD2, IInstructionOperandResolver {
 		public static ModuleDefMD Load(System.Reflection.Module mod, ModuleContext context, ImageLayout imageLayout) {
 			return Load(mod, new ModuleCreationOptions(context), imageLayout);
 		}
-
+
```
0xd4d: Extra spaces

src/DotNet/ModuleDefMD.cs
```diff
@@ -234,7 +234,20 @@ public sealed class ModuleDefMD : ModuleDefMD2, IInstructionOperandResolver {
...
 		}
-
+		
+#if NETSTANDARD2_0
+		//TODO: works only on windows
+		[DllImport("Kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+		static extern IntPtr GetModuleHandle(string lpModuleName);
+#endif
+		static IntPtr GetModuleHandle(System.Reflection.Module mod)
+		{
```
0xd4d: Should be on previous line

src/DotNet/ModuleDefMD.cs
```diff
@@ -234,7 +234,20 @@ public sealed class ModuleDefMD : ModuleDefMD2, IInstructionOperandResolver {
...
 			return Load(mod, new ModuleCreationOptions(context), imageLayout);
 		}
-
+		
+#if NETSTANDARD2_0
+		//TODO: works only on windows
+		[DllImport("Kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+		static extern IntPtr GetModuleHandle(string lpModuleName);
+#endif
+		static IntPtr GetModuleHandle(System.Reflection.Module mod)
```
0xd4d: `Marshal.GetHINSTANCE` exists on Windows and I think Mono also supports it. You could use reflection to get the method. If it doesn't exist, you could throw NotSupportedException

src/dnlib.netstandard.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
+        <RootNamespace>dnlib</RootNamespace>
+        <AssemblyName>dnlib</AssemblyName>
+        <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+        <SignAssembly>true</SignAssembly>
+        <AssemblyOriginatorKeyFile>..\dnlib.snk</AssemblyOriginatorKeyFile>
+        <OutputPath>../$(Configuration)/bin</OutputPath>
```
0xd4d: I'm pretty sure backslashes can still be used. MSBuild should convert them to forward slashes if needed

src/dnlib.netstandard.csproj
```diff
@@ -0,0 +1,21 @@
...
+        <DocumentationFile>$(OutputPath)/$(TargetFramework)/$(AssemblyName).xml</DocumentationFile>
```
0xd4d: Same here (backslashes)

src/DotNet/ModuleDefMD.cs
0xd4d: Only do this if it's a netstandard 2.0 build of course.

0xd4d:
> Dynamically created (in memory) module loading doesn't work at all.

What happens?

SlowLogicBoy:
>>Dynamically created (in memory) module loading doesn't work at all.

>What happens?

This is a limitation of `GetModuleHandle` but since I changed it to reflection, as you pointed out, it shouldn't matter anymore.

src/DotNet/ModuleDefMD.cs
```diff
@@ -235,6 +235,19 @@ public sealed class ModuleDefMD : ModuleDefMD2, IInstructionOperandResolver {
 			return Load(mod, new ModuleCreationOptions(context), imageLayout);
 		}
 
+		static IntPtr GetModuleHandle(System.Reflection.Module mod) {
+#if NETSTANDARD2_0
+			var methods = typeof(Marshal).GetMethods();
```
0xd4d: There should be a GetMethod(string, Type[]) you can call. You should verify that it takes a Module as its single parameter (in case they add overloads in the future)

0xd4d:
Thanks!

0xd4d:
Did I merge this too early. I noticed that you added two TODOs in your first post.

SlowLogicBoy:
Nah, I just made them as notes to check these on my linux machine. I will file issues if I'll find anything, but there shouldn't be any.

PR #158, opened 2018-02-21
Assembly Signing is not supported.

SlowLogicBoy:
Assembly signing is only supported on Windows with msbuild.

doesn't build on Linux with signing.
`dotnet build dnlib.netstandard.csproj` output
`CSC : error CS7027: Error signing output with public key from file '../dnlib.snk' -- Assembly signing not supported. [.../dnlib/src/dnlib.netstandard.csproj]`


0xd4d:
I think you need to do what @kiootic did [here](https://github.com/0xd4d/dnlib/commit/a4b2b89e85c059581919e4971a6d345de377fe4e)

See also https://github.com/dotnet/cli/issues/6911#issuecomment-309647478

0xd4d:
The first part only, not the FrameworkPathOverride part.

PR #160, opened 2018-02-21
Fix for #159

### LINKED ISSUE #159: netcoreapp2.0 Marshal.GetHINSTANCE on Linux platform
running on netcoreapp2.0.
```csharp
//static IntPtr GetModuleHandle(System.Reflection.Module mod)
//This is not null
var GetHINSTANCE = typeof(Marshal).GetMethod("GetHINSTANCE", new [] { typeof(System.Reflection.Module) });
if(GetHINSTANCE == null) 
    //doesn't throw
    throw new NotSupportedException("Module loading is not supported on current platform");
return (IntPtr)GetHINSTANCE.Invoke(null, new [] { mod }); //Returns -1; 


//public static ModuleDefMD Load(System.Reflection.Module mod, ModuleCreationOptions options, ImageLayout imageLayout)
if (addr == new IntPtr(-1))
    //throws here.
    throw new InvalidOperationException(string.Format("Module {0} has no HINSTANCE", mod)); 
```

Any actions needed?

SlowLogicBoy:
fixes #159 
this looks ugly...

### DIFF
--- a/src/DotNet/ModuleDefMD.cs
+++ b/src/DotNet/ModuleDefMD.cs
@@ -243,7 +243,7 @@ namespace dnlib.DotNet {
 
 			var addr = (IntPtr)GetHINSTANCE.Invoke(null, new [] { mod });
 			if(addr == new IntPtr(0) || addr == new IntPtr(-1))
-				throw new NotSupportedException("Module loading is not supported on current platform");
+				throw new ArgumentException("It is not possible to get address of module");
 
 			return addr;
 #else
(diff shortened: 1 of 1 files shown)


0xd4d:
ArgumentException is probably better. Should say that it's not possible to get the address of the module.

SlowLogicBoy:
FIY.
When using Mono (ether netstandard or netframework version of dnlib.dll) `GetHINSTANCE` doesn't work.

0xd4d:
Thanks for the testing and the fixes!

PR #178, opened 2018-04-09
Target both .NET Fx 3.5 and .NET Std 2.0 in dnlib.netstandard.csproj (closes #177)

### LINKED ISSUE #177: Target multiple frameworks in dnlib.netstandard.csproj
Changing `<TargetFramework>netstandard2.0</TargetFramework>` to `<TargetFrameworks>net35;netstandard2.0</TargetFrameworks>` (note the plural framework**s**) will generate assemblies targeting both frameworks on build. When adding dnlib as a project reference, it will allow the compiler to pick whichever framework type is most appropriate (for example net35 for a .NET Fx 3.5+ app, .NET Standard 2.0 for a .NET Core app) and potentially avoid adding 96 extra DLLs.

Personally, I can't use the old csproj file due to a VS bug, so I'm forced to use the .NET Standard one.

HoLLy-HaCKeR:
This change makes it so that the dnlib.netstandard.csproj file produces both .NET 3.5 and .NET Standard 2.0 binaries, and technically makes the original csproj obsolete (you need VS2017 for C# 7.2 anyway). This change also allows you to easily create a NuGet package on build that includes both versions using this csproj.

See issue #177. Also note that this isn't the Client version of .NET 3.5.

To build the .NET Standard version with .NET Core, use `dotnet build dnlib.netstandard.csproj -f netstandard2.0` now instead of `dotnet build dnlib.netstandard.csproj`.

### DIFF
--- a/src/dnlib.netstandard.csproj
+++ b/src/dnlib.netstandard.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <Import Project="$(SolutionDir)\DnlibCommon.props" Condition="Exists('$(SolutionDir)\DnlibCommon.props')" />
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>net35;netstandard2.0</TargetFrameworks>
     <RootNamespace>dnlib</RootNamespace>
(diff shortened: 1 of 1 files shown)


0xd4d:
Thanks!

PR #224, opened 2018-11-13
.NET Core: 'AppDomain' does not contain a definition for 'DefineDynamicAssembly'

KevOrr:
After a fresh clone, Visual Studio fails to build dnlib: `'AppDomain' does not contain a definition for 'DefineDynamicAssembly' and no accessible extension method 'DefineDynamicAssembly' accepting a first argument of type 'AppDomain' could be found (are you missing a using directive or an assembly reference?)` [MethodTableToTypeConverter.cs:33](https://github.com/0xd4d/dnlib/blob/6a52e50333486af740f098735dd39c0876e1fbee/src/DotNet/Emit/MethodTableToTypeConverter.cs#L33).

I think this was introduced in [this change](https://github.com/0xd4d/dnlib/commit/e4fe68d5b4d5776f1f8e737b88ee092f5df83040#diff-db76fc2e460c8d153e19af074d213089L30). [According to the docs](https://docs.microsoft.com/en-us/dotnet/standard/frameworks), the preprocessor directive to use is `NETCOREAPP2_0`, and indeed changing both occurrences of `NETCOREAPP` with `NETCOREAPP2_0` allows it to build.

### DIFF
--- a/src/DotNet/Emit/MethodTableToTypeConverter.cs
+++ b/src/DotNet/Emit/MethodTableToTypeConverter.cs
@@ -27,7 +27,7 @@ namespace dnlib.DotNet.Emit {
 
 		static MethodTableToTypeConverter() {
 			if (ptrFieldInfo == null) {
-#if NETSTANDARD2_0 || NETCOREAPP
+#if NETSTANDARD2_0 || NETCOREAPP2_0
 				var asmb = AssemblyBuilder.DefineDynamicAssembly(new AssemblyName("DynAsm"), AssemblyBuilderAccess.Run);
 #else
 				var asmb = AppDomain.CurrentDomain.DefineDynamicAssembly(new AssemblyName("DynAsm"), AssemblyBuilderAccess.Run);
--- a/src/PE/ProcessorArchUtils.cs
+++ b/src/PE/ProcessorArchUtils.cs
@@ -16,7 +16,7 @@ namespace dnlib.PE {
 		}
 
 		static class RuntimeInformationUtils {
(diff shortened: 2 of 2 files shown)


0xd4d:
I have no problem building it. What version of VS and .NET Core do you have installed?

KevOrr:
Visual Studio 2017 version 15.9.0

```
$ dotnet --info
.NET Command Line Tools (2.1.105)

Product Information:
 Version:            2.1.105
 Commit SHA-1 hash:  141cc8d976

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17134
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.105\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.7
  Build    : 2d61d0b043915bc948ebf98836fefe9ba942be11
```

0xd4d:
Seems like NETCOREAPP is only defined if you have .NET SDK 2.1 or later installed. Can you install that version and try again?

0xd4d:
Great! I'll close this because it's useful to be able to use NETCOREAPP instead of the one with a version number eg. NETCOREAPP2_0 since I don't have to update the source code after updating the target frameworks in the csproj file.

If anyone else is still using .NET Core SDK 2.0: MS has already stopped supporting .NET Core 2.0, and it's recommended to move to .NET Core 2.1 since it's the LTS release and currently the latest .NET Core release.

https://www.microsoft.com/net/platform/support-policy

PR #227, opened 2018-12-16
remap types in dnlib.DotNet.Importer

dyarkovoy:
I'm writing assembly processing library backed by dnlib and suggest a simple improvement to the dnlib.DotNet.Importer for the cases when the imported entity references renamed types in the target module. For example, type A has been imported into target assembly as A_1. 
Now, when processing a reference to A, Importer needs a way to remap the reference to A_1.

### DIFF
--- a/src/DotNet/Importer.cs
+++ b/src/DotNet/Importer.cs
@@ -53,7 +53,7 @@ namespace dnlib.DotNet {
 		/// </summary>
 		/// <param name="source"><see cref="ITypeDefOrRef"/> referenced by the entity that is being imported.</param>
 		/// <returns>matching <see cref="ITypeDefOrRef"/> or <c>null</c> if there's no match.</returns>
-		public virtual ITypeDefOrRef Map(ITypeDefOrRef source) => null;
+		public virtual TypeRef Map(ITypeDefOrRef source) => null;
 
 		/// <summary>
 		/// Matches source <see cref="FieldDef"/> to the one that is already present in the target module under a different name.
@@ -84,7 +84,7 @@ namespace dnlib.DotNet {
 	public struct Importer {
 		readonly ModuleDef module;
 		readonly GenericParamContext gpContext;
-		readonly ImportMapper Mapper;
+		readonly ImportMapper mapper;
 		RecursionCounter recursionCounter;
 		ImporterOptions options;
 
@@ -150,7 +150,7 @@ namespace dnlib.DotNet {
 			recursionCounter = new RecursionCounter();
 			this.options = options;
 			this.gpContext = gpContext;
-			Mapper = mapper;
+			this.mapper = mapper;
 		}
 
 		/// <summary>
@@ -660,9 +660,6 @@ namespace dnlib.DotNet {
 		public ITypeDefOrRef Import(TypeDef type) {
 			if (type == null)
 				return null;
-			var mapped = Mapper?.Map(type);
-			if (mapped != null)
-				return mapped;
 			if (TryToUseTypeDefs && type.Module == module)
 				return type;
 			return Import2(type);
@@ -673,6 +670,11 @@ namespace dnlib.DotNet {
 				return null;
 			if (!recursionCounter.Increment())
 				return null;
+			var mapped = mapper?.Map(type);
+			if (mapped != null) {
+				recursionCounter.Decrement();
+				return mapped;
+			}
 			TypeRef result;
 
 			var declType = type.DeclaringType;
@@ -707,18 +709,18 @@ namespace dnlib.DotNet {
 		/// </summary>
 		/// <param name="type">The type</param>
 		/// <returns>The imported type or <c>null</c></returns>
-		public ITypeDefOrRef Import(TypeRef type) {
-			var mapped = Mapper?.Map(type);
-			if (mapped != null)
-				return mapped;
-			return TryResolve(Import2(type));
-		}
+		public ITypeDefOrRef Import(TypeRef type) => TryResolve(Import2(type));
 
 		TypeRef Import2(TypeRef type) {
 			if (type == null)
 				return null;
 			if (!recursionCounter.Increment())
 				return null;
+			var mapped = mapper?.Map(type);
+			if (mapped != null) {
+				recursionCounter.Decrement();
+				return mapped;
+			}
 			TypeRef result;
 
 			var declaringType = type.DeclaringType;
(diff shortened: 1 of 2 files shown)


src/DotNet/Importer.cs
```diff
@@ -43,6 +43,13 @@ public enum ImporterOptions {
 		FixSignature = int.MinValue,
 	}
 
+	/// <summary>
+	/// Re-maps entities that were renamed in the target module
+	/// </summary>
+	public interface IImportMapper {
+		ITypeDefOrRef Map(ITypeDefOrRef source);
```
0xd4d: Missing XML doc comments

src/DotNet/Importer.cs
```diff
@@ -52,6 +59,7 @@ public struct Importer {
 		readonly GenericParamContext gpContext;
 		RecursionCounter recursionCounter;
 		ImporterOptions options;
+		public IImportMapper Mapper;
```
0xd4d: It should be readonly and not public. Add a new ctor with an extra IImportMapper arg.

src/DotNet/Importer.cs
```diff
@@ -613,6 +622,8 @@ public Importer(ModuleDef module, ImporterOptions options)
 		public ITypeDefOrRef Import(TypeDef type) {
 			if (type == null)
 				return null;
+			var mapped = Mapper?.Map(type);
+			if (mapped != null) return mapped;
```
0xd4d: nit: put the return on the next line. Same below in the other method.

src/DotNet/Importer.cs
```diff
@@ -657,7 +668,11 @@ public Importer(ModuleDef module, ImporterOptions options)
 		/// </summary>
 		/// <param name="type">The type</param>
 		/// <returns>The imported type or <c>null</c></returns>
-		public ITypeDefOrRef Import(TypeRef type) => TryResolve(Import2(type));
+		public ITypeDefOrRef Import(TypeRef type) {
```
0xd4d: The TypeSpec overload should also call Map()

dyarkovoy:
Agree on everything, have question about TypeSpec overload, doesn't it eventually call Import(ITypeDef type) or  Import(ITypeRef type) ? I see the following route: Import(TypeSpec) -> CreateClassOrValueType(ITypeDefOrRef)  -> Import(ITypeDefOrRef) -> Import(IType) -> Import(TypeDef or TypeRef), the last one is being mapped already.
Am I missing something here?

I'm 100% for the generic absract class with mappers for the members, will prepare a new PR shortly

0xd4d:
Yes forget about that, it's probably not very useful to return your own TypeSpec.

BTW, you don't need to create a new PR, just push to the same branch.

dyarkovoy:
made the changes, let me know if anything else is needed to get it merged

src/DotNet/Importer.cs
```diff
@@ -971,6 +1026,9 @@ public Importer(ModuleDef module, ImporterOptions options)
 				return field;
 			if (!recursionCounter.Increment())
 				return null;
+			var mapped = Mapper?.Map(field);
+			if (mapped != null)
+				return mapped;
```
0xd4d: Needs to call recursionCounter.Decrement().

src/DotNet/Importer.cs
```diff
@@ -1000,6 +1058,9 @@ public Importer(ModuleDef module, ImporterOptions options)
 				return method;
 			if (!recursionCounter.Increment())
 				return null;
+			var mapped = Mapper?.Map(method);
+			if (mapped != null)
+				return mapped;
```
0xd4d: Same here

src/DotNet/Importer.cs
```diff
@@ -1037,6 +1098,9 @@ public Importer(ModuleDef module, ImporterOptions options)
 				return null;
 			if (!recursionCounter.Increment())
 				return null;
+			var mapped = Mapper?.Map(memberRef);
+			if (mapped != null)
+				return mapped;
```
0xd4d: And here

src/DotNet/Importer.cs
```diff
@@ -613,6 +660,9 @@ public Importer(ModuleDef module, ImporterOptions options)
 		public ITypeDefOrRef Import(TypeDef type) {
 			if (type == null)
 				return null;
+			var mapped = Mapper?.Map(type);
```
0xd4d: This code must be in Import2()

src/DotNet/Importer.cs
```diff
@@ -657,7 +707,12 @@ public Importer(ModuleDef module, ImporterOptions options)
 		/// </summary>
 		/// <param name="type">The type</param>
 		/// <returns>The imported type or <c>null</c></returns>
-		public ITypeDefOrRef Import(TypeRef type) => TryResolve(Import2(type));
+		public ITypeDefOrRef Import(TypeRef type) {
+			var mapped = Mapper?.Map(type);
```
0xd4d: This code must be in Import2()

src/DotNet/Importer.cs
```diff
@@ -43,13 +43,48 @@ public enum ImporterOptions {
...
+	}
+
 	/// <summary>
 	/// Imports <see cref="Type"/>s, <see cref="ConstructorInfo"/>s, <see cref="MethodInfo"/>s
 	/// and <see cref="FieldInfo"/>s as references
 	/// </summary>
 	public struct Importer {
 		readonly ModuleDef module;
 		readonly GenericParamContext gpContext;
+		readonly ImportMapper Mapper;
```
0xd4d: Should be lower case: `mapper`

dyarkovoy:
Right, I overlooked RecusionCounter. Actually, I would refactor RecursionCounter.Increment() to return IDisposable so we can wrap recursion-prone chunks in the using(recursionCounter.Increment()) { } blocks and avoid having to call Decrement() altogether. Let me know if you like the idea and will do the needful, or you just want me to add missing Decrement()'s

0xd4d:
No refactoring is needed at this time.

0xd4d:
Unless you have something else to push, I'm ready to merge this.

0xd4d:
You can restore that change so they're in Import() and not the recursive Import2().

dyarkovoy:
The problem with having it in Import2 is that mapping TypeDefs to TypeRefs introduces additional (huge!) overhead of resolving TypeRefs back to TypeDefs, as it uses very slow TypeDefFinder.Find()
My code that was taking 8 seconds previously, has now been running for 20 minutes and is like only 1% finished.
I reversed the change and you can merge now. 
I'm not sure this is the perfect fix though, it may be better to split Map into TypeDef and TypeRef versions, or use cache in TypeDefFinder, will have to further research this, but for now we can merge.

0xd4d:
TypeDefFinder can cache the result and it will be faster, but check the docs first since you have to disable it if you eg. rename types (but you can enable it afterwards)

Thanks for the PR, merging now.

Issue #244, opened 2019-02-09
Customize importing of System.Type's


dyarkovoy:
Suppose we run from NetCore app and want to inject some dynamically generated code into NETFramework assembly. Let's say we want to reference System.Text.StringBuilder so we do dnlib.DotNet.Importer.Import(typeof(System.Text.StringBuilder)), and it returns TypeRef with reference to System.Private.CoreLib. We can't inject it as is, because our target assembly is NETFramework, 
so we need to change System.Private.CoreLib to mscorlib. Do you think it makes sense to check in the Importer.CreateScopeReference(Type type) whether the type is from corlib and use module.CorLibTypes.AssemblyRef, or is there a better solution to this?

0xd4d:
No, the solution is to not use runtime types, always reference the reference types and it will just work.

dyarkovoy:
Okay, I believe I understand what you mean, but not sure I fully understand how to do it. You mean these stub assemblies that contain only TypeForwardedTo() attributes and located under %ProgramFiles(x86)%\Reference Assemblies, right? I wasn't able to find complete set on NuGet (assembly containing System.Text.StringBuilder in particular), so I just downloaded them from Microsoft.
So now before I import any System.Type or MemberInfo via Importer, I must load reference assembly for the corresponding target at runtime, resolve the type/member via reference assembly, and then only pass it to Importer, is that what you mean?

0xd4d:
You shouldn't import any runtime types since they're in runtime assemblies and not reference assemblies. If you need StringBuilder, you have to find it in the reference assembly and import it. You can get the locations of the assemblies the C# compiler uses by enabling diagnostic output when building a simple hello world app. (VS -> Options -> Projects and Solutions -> Build and Run -> MSBuild project build output verbosity -> detailed). Your assembly resolver should be updated to only return reference assemblies.

dyarkovoy:
I see your point. In this case, how about adding 

> public virtual TypeRef Map(Type source) => null;

to ImportMapper and calling it from CreateTypeRef2(Type type) ?

I would then search for the reference assemblies in the mapper, instead of doing it for every type and member before calling Import

0xd4d:
Yes, but call it from `CreateTypeRef(Type type)` so it does the same thing `Import(TypeDef type)` does.

dyarkovoy:
Done, PR #245

FenixDan:
Could you give an example of use for this solution?

[...]

PR #245, opened 2019-02-11
Customize importing of System.Type's

dyarkovoy:
see issue #244


0xd4d:
This isn't the correct way to solve it. See my reply in the issue you created.

PR #281, opened 2019-06-23
Allow overwriting of the dll name which should be imported

batzen:
This PR adds the ability to overwrite the dll name being used/imported in `ImportDirectory`.

The purpose/use case for this is that .NET core libraries still have "mscoree.dll" in that section but to enable the DLLExport scenario for .NET core we have to use something different to be able to inject our own host/runtime resolution logic.
I use this technique to replace "mscoree.dll" with "ijwhost.dll". I guess the normal compile of .NET core will, sooner or later, also replace "mscoree.dll" in that place with something else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,7 @@ Exporting managed methods (DllExport)
 -------------------------------------
 
 dnlib supports exporting managed methods so the managed DLL file can be loaded by native code and then executed. .NET Framework supports this feature, but there's no guarantee that other CLRs (eg. .NET Core or Mono/Unity) support this feature.
+In case of .NET Core please be aware that `ijwhost.dll` has to be loaded prior to calling your exported method and that ijwhost currently (as of .NET Core 3.0) does not work if the calling app is self-contained.
 
 The `MethodDef` class has an `ExportInfo` property. If it gets initialized, the method gets exported when saving the module. At most 65536 (2^16) methods can be exported. This is a PE file limitation, not a dnlib limitation.
 
--- a/src/DotNet/Writer/ImportDirectory.cs
+++ b/src/DotNet/Writer/ImportDirectory.cs
@@ -16,7 +16,7 @@ namespace dnlib.DotNet.Writer {
 		uint length;
 		RVA importLookupTableRVA;
 		RVA corXxxMainRVA;
-		RVA mscoreeDllRVA;
+		RVA dllToImportRVA;
 		int stringsPadding;
 
 		/// <summary>
@@ -50,6 +50,11 @@ namespace dnlib.DotNet.Writer {
 
 		internal bool Enable { get; set; }
 
+		/// <summary>
+		/// Gets/sets the name of the dll which should be imported.
+		/// </summary>
+		public string DllToImport { get; set; } = "mscoree.dll";
+
 		const uint STRINGS_ALIGNMENT = 16;
 
 		/// <summary>
@@ -71,7 +76,7 @@ namespace dnlib.DotNet.Writer {
 			length += (uint)stringsPadding;
 			corXxxMainRVA = rva + length;
 			length += 0xE;
-			mscoreeDllRVA = rva + length;
+			dllToImportRVA = rva + length;
 			length += 0xC;
 			length++;
 		}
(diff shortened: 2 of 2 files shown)


0xd4d:
Is `ijwhost.dll` an official .NET Core 3.0 binary, and is it documented that this is supported/working? If so, the README.md should also be updated. It would also be better if dnlib detected this so it could initialize that property itself instead of having to rely on the user to remember to do that.

batzen:
`ijwhost.dll` is the name that's currently being used in .NET Core 3.0.
You can read about it at https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/host-components.md
I don't know if it stays that way. And even a compile on .NET Core 3.0 currently emits `mscoree.dll`.
I will create a PR in dotnet changing that name or at least propose it to be changed as it will never work if that name stays the same.

I fully agree with you that dnlib should take care of choosing the right name automatically in the future, but being able to overwrite that name opens up further possibilities like providing your own DLL if you want to do something special to initialize the runtime when your DLL gets loaded.

I will create another PR in core-setup to get IJW support working correctly, but if MS rejects my PR or decides stay with `mscoree.dll` it would be nice to be able to overwrite this with dnlib.

According to the readme:
I will create a PR here with updated instructions on how to get DLLExport running in .NET Core 3.0 once my changes got merged in dotnet itself or if it got rejected, as the instructions vary quite a bit depending on how many (if any) of my changes will get merged in dotnet itself.

batzen:
Any chance of getting this merged?

0xd4d:
> Any chance of getting this merged?

Does this work on .NET Core 3.0?

Like said above, the README should be updated too.

0xd4d:
Are you still workign on this?

batzen:
Would updating the readme be enough for you to accept the PR or should i add code to auto detect the required dll name?

0xd4d:
No, the readme should be enough.

Does it work on .NET Core 3.0?

batzen:
It does if ijwhost is loaded and it's not a self contained app. That's what i'll also add to the readme.

batzen:
So i am not that familiar with the code, but i guess i also have to adjust `length += 0xC;` to be `length += this.DllToImport.Length + 1;` or?
In my use case it worked because "mscoree.dll\0" has the same length as "ijwhost.dll\0".
And then we should also change `length += 0xE;` to be `length += this.EntryPointName.Length + 1;`, while also introducing 
```
public string EntryPointName  {
	get => entryPointName ?? (IsExeFile ? "_CorExeMain" : "_CorDllMain");
	set => entryPointName = value;
}
```

And we should do the same for `DllToImport`. That way we can add auto detection for different frameworks later and users could still overwrite the values if they want to.
What do you think?

batzen:
Looking at the code i am wondering what `length = 0x28;` stands for.
Why is it hard coded and what does it mean?

0xd4d:
`0x28` is the first 0x28 bytes of the data, which you can see in `WriteTo()`

You're right that length += 0xC and +0xE refer to the zero terminated strings, and at least the 0xC line needs to be updated. The 0xE line doesn't need to be updated, since the strings have identical length but it doesn't hurt either to make it more obvious so feel free to update it too if you want.

0xd4d:
Note that the 0xE line is actually string len + zero + 2 bytes

batzen:
> 0x28 is the first 0x28 bytes of the data, which you can see in WriteTo()

I got that, but what does it stand for? Why exactly 0x28? Does it have some special meaning? If it does we should give it a name. ;-)

I would replace `0xE` with `length += (uint)EntryPointName.Length + ZERO_STRING_TERMINATION + TWO_BYTE_PADDING;` then.
That should make it very clear.

0xd4d:
Like I wrote, 0x28 is the number of bytes written in the method before the other data (eg. strings).

There's no need for a 'const'.

batzen:
I just noticed that everything we need to automatically use ijwhost is the following at the bottom of `ModuleWriter.CreateChunks`
```c#
module.Assembly.TryGetOriginalTargetFrameworkAttribute(out var framework, out var version, out var profile);

if (framework == ".NETCoreApp")
{
	importDirectory.DllToImport = "ijwhost.dll";
}
```
Should i add that too?
Then the only part i'd have to add to the readme would be about having ijwhost loaded before calling an exported method on .net core.

0xd4d:
No, the default strings should be used since all tools use e.g. `mscoree.dll`. If someone wants to use ijwhost.dll, they have to change it themselves.

0xd4d:
Thanks, merged.

I will undo your 'const' updates though since it doesn't make the code easier to read. Eg. most people should know that a zero-terminator is 1 byte.

batzen:
That's not the reason i use a const. It was to clarify why 1 was added. I first tried a comment after 1, but it looked awful. Especially for the case where a padding was also needed.

PR #343, opened 2020-03-11
Improve IsPinvokeMethod

CreateAndInject:
.Net allows there're some spaces in the end of module name (`\t` `\n` aren't allowed)

By the way, if someone's PR is accepted, and then the original author submit a new commit to modify a little code, how to update to the lastest version?

Last time, I send a PR to dnlib and be accepted, and then you submit a new commit to remove a `?`, today, when I want to update to the lastest dnlib as following:
1. Click `New pull request`
2. 0xd4d/dnlib master <- CreateAndInject/dnlib master
The button `Merge pull request` is disabled, and I can't merge pull request since there's a conflict:
image

I think the conflict it's because you submit a new commit to remove a `?`.
So how can I resolve the conflict and merge the pull request without any extra tools installed on my pc? (such as github desktop and git cmd, just do it in www.github.com or VS2019)
Now, I can only delete the repo and fork again.

### DIFF
--- a/src/DotNet/ICodedToken.cs
+++ b/src/DotNet/ICodedToken.cs
@@ -601,6 +601,13 @@ namespace dnlib.DotNet {
 		public static int GetParamCount(this IMethod method) => method?.MethodSig.GetParamCount() ?? 0;
 
 		/// <summary>
+		/// Checks whether any normal visible parameter exists, doesn't include the hidden 'this' parameter
+		/// </summary>
+		/// <param name="method">this</param>
+		/// <returns><c>true</c> if there's at least one normal visible parameter</returns>
+		public static bool HasParams(this IMethod method) => method.GetParamCount() > 0;
+
+		/// <summary>
 		/// Gets a normal visible parameter, doesn't include the hidden 'this' parameter
 		/// </summary>
 		/// <param name="method">this</param>
--- a/src/DotNet/ImplMap.cs
+++ b/src/DotNet/ImplMap.cs
@@ -2,6 +2,8 @@
 
 using System;
 using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
 using dnlib.DotNet.MD;
 
 namespace dnlib.DotNet {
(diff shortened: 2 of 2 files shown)


src/DotNet/ICodedToken.cs
```diff
@@ -600,6 +600,13 @@ public static partial class Extensions {
 		/// <returns>Normal visible parameter count</returns>
 		public static int GetParamCount(this IMethod method) => method?.MethodSig.GetParamCount() ?? 0;
 
+		/// <summary>
+		/// Checks whether any normal visible parameter exists, doesn't include the hidden 'this' parameter
+		/// </summary>
+		/// <param name="method">this</param>
+		/// <returns><c>true</c> if there's at least one normal visible parameter</returns>
+		public static bool HasParam(this IMethod method) => method.GetParamCount() > 0;
```
0xd4d: `HasParams`

CreateAndInject:
> About your question: you should always create a new branch with your changes and not use the `master` branch.

I got it, thanks!

src/DotNet/ImplMap.cs
```diff
@@ -215,6 +215,7 @@ public abstract class ImplMap : IMDTokenProvider {
 		}
 
 		static string GetDllName(string dllName) {
+			dllName = dllName.TrimEnd(' ');
```
0xd4d: Your post mentions that all whitespace is ignored. You can pass in `null` here and all whitespace will be trimmed.

CreateAndInject:
No, I say: `\t` `\n` aren't allowed.

This work fine:
```
[DllImport("user32   ")]
static extern IntPtr GetDesktopWindow();
```

DllNotFoundException:
```
[DllImport("user32\t")]
static extern IntPtr GetDesktopWindow();
```

0xd4d:
Have you looked at the coreclr source code and confirmed that only `' '` is ignored at the end of the string?

If so you should create a `static readonly` array that you pass to `TrimEnd()` so it doesn't allocate an array every time it's called.

CreateAndInject:
I didn't view .net source, I just usually meet DllImport with a space in the end (I think it's probably because the code is copied from web page), and the program work fine, but IsPinvokeMethod failed.
So I made a test: space in the begin are not allowed, `\t` `\n` are not allowed whatever in the begin or end.

CreateAndInject:
I think it's not matter whehter to create a static readonly array, `IsPinvokeMethod` isn't called by dnlib, and it's less called in our program.

0xd4d:
It shouldn't allocate if it can avoid it so you need a static readonly array.

The code should also only ignore the spaces if it's Windows because this looks like a Windows OS feature.

0xd4d:
This function needs a new bool param so you can force Windows behaviour if it's run on Linux.

CreateAndInject:
GetDllName is private and isn't used directly.

User use IsPinvokeMethod to check, it's property, so how do you want to impl it?

CreateAndInject:
Any problem?

0xd4d:
Quick checks:

- it's a breaking change to add a new parameter.
- The Windows check is probably wrong. Also it shouldn't force windows behavior if it's Windows, the forceWindows option overrides everything.

CreateAndInject:
>The Windows check is probably wrong.
How to check?
1. There's a warning(MaxOSX, Win32S, Win32Windows, WinCE, Xbox), but no compiler erros.
2. Is Xbox a version of Windows?
image

CreateAndInject:
https://docs.microsoft.com/en-us/dotnet/api/system.platformid
The possible value is MacOSX, Unix, Win32NT, and Win32S, Win32Windows, WinCE, Xbox are no longer in use.

So I think my lastest commit `Adjust` can work.

0xd4d:
Try this instead:

```C#
		static bool IsWindows() =>
#if NETSTANDARD
			RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
#else
			Path.DirectorySeparatorChar == '\\' || Path.AltDirectorySeparatorChar == '\\';
#endif
```

CreateAndInject:
It's OK now.

0xd4d:
Thanks!

PR #362, opened 2020-06-11
Update ModuleDefMD.cs

CursedSheep:
cache system for string type operand


0xd4d:
I was just going to comment on this, I guess you don't need the fix anymore.

PR #378, opened 2020-07-06
Support for custom/experimental CIL opcodes.

### LINKED ISSUE #367: Support for custom CIL instructions
Hi,

For a project I'm currently working on, I need to be able to insert custom CIL instructions into a method body (ones that are understood by the runtime I'm targeting). Basically, I just want to be able to provide dnlib with custom `OpCode` instances that inform it about how to decode/encode these instructions, calculate stack usage and control flow, etc.

From what I can gather looking at the source, what would need to happen for this to be possible is:

* `OpCode` would need to be made extensible in some form (probably just a public constructor).
* There would need to be some place in the API for registering custom `OpCode` instances.
* `MethodBodyReaderBase` and a few other places need to be made aware about the experimental opcode range (`0xf0`..`0xfb`).

There may of course be other required changes that I'm not aware of, but I think that's the gist of it.

Would a patch for this be welcome? If so, what would be an acceptable design for this?

alexrp:
Closes #367.

### DIFF
--- a/src/DotNet/Emit/DynamicMethodBodyReader.cs
+++ b/src/DotNet/Emit/DynamicMethodBodyReader.cs
@@ -149,7 +149,8 @@ namespace dnlib.DotNet.Emit {
 		/// instance or a DynamicResolver instance.</param>
 		/// <param name="importer">Importer</param>
 		/// <param name="options">Options</param>
-		public DynamicMethodBodyReader(ModuleDef module, object obj, Importer importer, DynamicMethodBodyReaderOptions options) {
+		public DynamicMethodBodyReader(ModuleDef module, object obj, Importer importer, DynamicMethodBodyReaderOptions options)
+			: base(module.Context) {
 			this.module = module;
 			this.importer = importer;
 			this.options = options;
--- a/src/DotNet/Emit/MethodBodyReaderBase.cs
+++ b/src/DotNet/Emit/MethodBodyReaderBase.cs
@@ -25,7 +25,7 @@ namespace dnlib.DotNet.Emit {
 		/// <summary>Start offset of method</summary>
 		protected uint codeStartOffs;
 		/// <summary>The module context</summary>
-		protected ModuleContext context;
+		protected readonly ModuleContext context;
 
 		/// <summary>
 		/// Gets all parameters
--- a/src/DotNet/Emit/MethodBodyReaderBase.cs
+++ b/src/DotNet/Emit/MethodBodyReaderBase.cs
@@ -24,6 +24,8 @@ namespace dnlib.DotNet.Emit {
 		protected uint codeEndOffs;
 		/// <summary>Start offset of method</summary>
 		protected uint codeStartOffs;
+		/// <summary>The module context</summary>
+		protected ModuleContext context;
 
 		/// <summary>
 		/// Gets all parameters
@@ -54,6 +56,14 @@ namespace dnlib.DotNet.Emit {
 		/// <summary>
 		/// Constructor
 		/// </summary>
+		/// <param name="context">The module context</param>
+		protected MethodBodyReaderBase(ModuleContext context) {
+			this.context = context;
+		}
+
+		/// <summary>
+		/// Constructor
+		/// </summary>
 		/// <param name="reader">The reader</param>
 		protected MethodBodyReaderBase(DataReader reader)
 			: this(reader, null) {
@@ -64,9 +74,20 @@ namespace dnlib.DotNet.Emit {
 		/// </summary>
 		/// <param name="reader">The reader</param>
 		/// <param name="parameters">Method parameters or <c>null</c> if they're not known yet</param>
-		protected MethodBodyReaderBase(DataReader reader, IList<Parameter> parameters) {
+		protected MethodBodyReaderBase(DataReader reader, IList<Parameter> parameters)
+			: this(reader, parameters, null) {
+		}
+
+		/// <summary>
+		/// Constructor
+		/// </summary>
+		/// <param name="reader">The reader</param>
+		/// <param name="parameters">Method parameters or <c>null</c> if they're not known yet</param>
+		/// <param name="context">The module context</param>
+		protected MethodBodyReaderBase(DataReader reader, IList<Parameter> parameters, ModuleContext context) {
 			this.reader = reader;
 			this.parameters = parameters;
+			this.context = context;
 		}
 
 		/// <summary>
@@ -218,9 +239,11 @@ namespace dnlib.DotNet.Emit {
 		/// </summary>
 		OpCode ReadOpCode() {
 			var op = reader.ReadByte();
-			if (op != 0xFE)
-				return OpCodes.OneByteOpCodes[op];
-			return OpCodes.TwoByteOpCodes[reader.ReadByte()];
+			if (op == 0xFE)
+				return OpCodes.TwoByteOpCodes[reader.ReadByte()];
+			if (op >= 0xF0 && op <= 0xFB)
+				return context?.GetExperimentalOpCode(op, reader.ReadByte()) ?? OpCodes.UNKNOWN2;
+			return OpCodes.OneByteOpCodes[op];
 		}
 
 		/// <summary>
--- a/src/DotNet/Emit/MethodBodyReaderBase.cs
+++ b/src/DotNet/Emit/MethodBodyReaderBase.cs
@@ -24,8 +24,7 @@ namespace dnlib.DotNet.Emit {
 		protected uint codeEndOffs;
 		/// <summary>Start offset of method</summary>
 		protected uint codeStartOffs;
-		/// <summary>The module context</summary>
-		protected readonly ModuleContext context;
+		readonly ModuleContext context;
 
 		/// <summary>
 		/// Gets all parameters
--- a/src/DotNet/Emit/MethodBodyReaderBase.cs
+++ b/src/DotNet/Emit/MethodBodyReaderBase.cs
@@ -240,8 +240,12 @@ namespace dnlib.DotNet.Emit {
 			var op = reader.ReadByte();
 			if (op == 0xFE)
 				return OpCodes.TwoByteOpCodes[reader.ReadByte()];
-			if (op >= 0xF0 && op <= 0xFB)
-				return context?.GetExperimentalOpCode(op, reader.ReadByte()) ?? OpCodes.UNKNOWN2;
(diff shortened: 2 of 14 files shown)


src/DotNet/Emit/DynamicMethodBodyReader.cs
```diff
@@ -207,6 +207,7 @@ public DynamicMethodBodyReader(ModuleDef module, object obj, Importer importer)
 			reader = ByteArrayDataReaderFactory.CreateReader(code);
 			method = CreateMethodDef(delMethod);
 			parameters = method.Parameters;
+			context = module.Context;
```
0xd4d, line 210: It should be able to pass this to the base ctor.

src/DotNet/Emit/MethodBodyReaderBase.cs
```diff
@@ -24,6 +24,8 @@ public abstract class MethodBodyReaderBase {
 		protected uint codeEndOffs;
 		/// <summary>Start offset of method</summary>
 		protected uint codeStartOffs;
+		/// <summary>The module context</summary>
+		protected ModuleContext context;
```
0xd4d, line 28: This should be readonly. I'd also rename the field to `contextOrNull` since I currently don't use NRT in dnlib.

src/DotNet/Emit/MethodBodyReaderBase.cs
```diff
@@ -218,9 +231,14 @@ protected MethodBodyReaderBase(DataReader reader)
 		/// </summary>
 		OpCode ReadOpCode() {
 			var op = reader.ReadByte();
-			if (op != 0xFE)
-				return OpCodes.OneByteOpCodes[op];
-			return OpCodes.TwoByteOpCodes[reader.ReadByte()];
+			if (op == 0xFE)
+				return OpCodes.TwoByteOpCodes[reader.ReadByte()];
+			if (op >= 0xF0 && op <= 0xFB) {
+				var op2 = reader.ReadByte();
```
0xd4d, line 237: You should only read this if `TryGetValue()` is successful.

src/DotNet/ModuleContext.cs
```diff
@@ -9,6 +11,7 @@ namespace dnlib.DotNet {
 	public class ModuleContext {
 		IAssemblyResolver assemblyResolver;
 		IResolver resolver;
+		Dictionary<byte, OpCode[]> experimentalOpCodes;
```
0xd4d, line 14: I think an `OpCode[][]` is better. We know the exact size of it and it's small. The lookups will be very fast as well.

Lazy initialization shouldn't be needed, it's small so you can allocate in the ctor and set the field to `readonly`.

The `ExperimentalOpCodes` property would then need to be removed and replaced with a method that returns the opcode (`TryGetExperimentalOpCode()`)

src/DotNet/Writer/MethodBodyWriterBase.cs
```diff
@@ -12,6 +12,8 @@ public abstract class MethodBodyWriterBase {
 		protected IList<Instruction> instructions;
 		/// <summary/>
 		protected IList<ExceptionHandler> exceptionHandlers;
+		/// <summary>The module context</summary>
+		protected ModuleContext context;
```
0xd4d, line 16: Should be `readonly` and renamed to `contextOrNull`.

Is this field used anywhere? I couldn't find any uses.

src/DotNet/Writer/MethodBodyWriterBase.cs
```diff
@@ -158,20 +172,19 @@ public abstract class MethodBodyWriterBase {
...
-				writer.WriteByte((byte)code);
-			}
-			else if (code == Code.UNKNOWN1)
-				writer.WriteByte((byte)Code.Nop);
-			else if (code == Code.UNKNOWN2)
-				writer.WriteUInt16((ushort)(((ushort)Code.Nop << 8) | Code.Nop));
-			else {
-				Error("Unknown instruction");
-				writer.WriteByte((byte)Code.Nop);
+			if (instr.OpCode.Size == 2)
```
0xd4d, line 175: This whole method needs to be reformatted (braces on the wrong line, `else` on the same line as `{`)

I also don't like that the whole method got rewritten. Is it not possible to just add the minimal amount of code here to support the new feature without rewriting the whole method?

0xd4d:
Thanks, just a few changes needed then I think this can be merged!

src/DotNet/Emit/DynamicMethodBodyReader.cs
alexrp: I'm just following the existing code here. I'm not sure why several of these fields (`reader`, `parameters`, etc) are being set in this way rather than passed to the base constructor.

src/DotNet/Emit/MethodBodyReaderBase.cs
alexrp: The lack of `readonly` is due to the change you commented on above. Several of these fields should ideally be `readonly` in my mind, but I assumed there's some reason why things are the way they are.

src/DotNet/Writer/MethodBodyWriterBase.cs
alexrp: Looks like a leftover. Shouldn't be necessary.

src/DotNet/Writer/MethodBodyWriterBase.cs
alexrp: I just rewrote it like this because it seemed kind of unwieldy to me after adding the case for experimental opcodes, and it occurred to me that it makes more sense to check the size rather than special-casing the opcode ranges.

It's not something I feel particularly strongly about. I can retain the old logic if you prefer.

src/DotNet/Emit/DynamicMethodBodyReader.cs
0xd4d: It's because it has to extract various things through reflection and it can't do that in a base() call. It would've been better to turn it into some Create() method instead which does all of that and then calls some internal ctor but it's too late now.

The context can easily be passed to the base, and that will also allow us to make the field readonly.

src/DotNet/Emit/MethodBodyReaderBase.cs
0xd4d: It's a breaking change to change the other ones to readonly now but this new field can be made readonly.

src/DotNet/ModuleContext.cs
alexrp: So just to be clear, are you suggesting an array of size 12 (to which we map the range `0xf0`..`0xfb`) where each element is an array of 256 `OpCode` references?

0xd4d: Yes an array of size 12 where each element is null. Initialization of this array should be done by the app before module is opened so Interlocked/etc isn't needed.

0xd4d:
Are you still working on this?

alexrp:
I've been side-tracked with other things recently. I'll try to get to it next week.

alexrp:
@0xd4d hey, sorry for the delay, I ended up being busy with other things for a while. I think I've addressed all the PR comments.

0xd4d:
It's better if you do not force push because now I have to re-read the whole thing instead of just the actual changes you made.

I had a very quick look but there are two comments left that haven't been addressed, the readonly comment and restoring the old code so it's easier to verify that it's still doing the same thing.

src/DotNet/Writer/MethodBodyWriterBase.cs
0xd4d: Seems like it's still left in the new code?

alexrp:
> It's better if you do not force push because now I have to re-read the whole thing instead of just the actual changes you made.

Sorry, I sometimes forget GitHub has an actual rebase merge feature on PRs these days.

> the readonly comment

I made the changes necessary for that one but forgot to actually go back and apply the modifier. 😕

> and restoring the old code so it's easier to verify that it's still doing the same thing.

Done.

src/DotNet/Writer/MethodBodyWriterBase.cs
```diff
@@ -12,6 +12,8 @@ public abstract class MethodBodyWriterBase {
 		protected IList<Instruction> instructions;
 		/// <summary/>
 		protected IList<ExceptionHandler> exceptionHandlers;
+		/// <summary>The module context</summary>
+		protected readonly ModuleContext context;
```
0xd4d, line 16: Does this need to be protected?

Also no code uses it so it seems like it's not needed.

0xd4d (changes requested):

src/DotNet/Emit/MethodBodyReaderBase.cs
```diff
@@ -218,9 +239,11 @@ protected MethodBodyReaderBase(DataReader reader)
 		/// </summary>
 		OpCode ReadOpCode() {
 			var op = reader.ReadByte();
-			if (op != 0xFE)
-				return OpCodes.OneByteOpCodes[op];
-			return OpCodes.TwoByteOpCodes[reader.ReadByte()];
+			if (op == 0xFE)
+				return OpCodes.TwoByteOpCodes[reader.ReadByte()];
+			if (op >= 0xF0 && op <= 0xFB)
+				return context?.GetExperimentalOpCode(op, reader.ReadByte()) ?? OpCodes.UNKNOWN2;
```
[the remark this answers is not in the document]
alexrp: What do you think about:

```csharp
		OpCode ReadOpCode() {
			var op = reader.ReadByte();
			if (op == 0xFE)
				return OpCodes.TwoByteOpCodes[reader.ReadByte()];
			if (op >= 0xF0 && op <= 0xFB && reader.BytesLeft >= 1) {
				if (context?.GetExperimentalOpCode(op, reader.ReadByte()) is OpCode opCode)
					return opCode;
				else
					reader.Position--;
			}
			return OpCodes.OneByteOpCodes[op];
		}
```

That would retain the old behavior in all cases except when there's a 2nd byte available in the stream and there's an actual custom opcode match. Either is fine with me (it doesn't matter for my use case) but I imagine this would be the least surprising from the average user's point of view.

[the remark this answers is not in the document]
0xd4d: Yes that works too and will do almost the same thing (`GetExperimentalOpCode()` can return `null` even if the feature is enabled by the user in which case the code will now return a 1-byte opcode).

alexrp:
@0xd4d I think that should cover everything.

0xd4d (approved):

0xd4d:
Thanks, it's now merged!

PR #382, opened 2020-07-19
Unescape Type.Name & Type.FullName, fix #370

### LINKED ISSUE #370: Import(System.Type) should unescape

CreateAndInject:
No need to unescape Type.Namespace.
This PR can fix #370

### DIFF
--- a/src/DotNet/ReflectionExtensions.cs
+++ b/src/DotNet/ReflectionExtensions.cs
@@ -108,7 +108,7 @@ namespace dnlib.DotNet {
 			!(type is null) && !type.HasElementType && (!type.IsGenericType || type.IsGenericTypeDefinition);
 
 		internal static string Unescape(string name) {
-			if (string.IsNullOrEmpty(name) || !name.Contains(@"\"))
+			if (string.IsNullOrEmpty(name) || name.IndexOf('\\') < 0)
 				return name;
 			var sb = new StringBuilder(name.Length);
 			for (int i = 0; i < name.Length; i++) {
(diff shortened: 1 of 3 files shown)


src/DotNet/ReflectionExtensions.cs
```diff
@@ -105,5 +105,11 @@ static class ReflectionExtensions {
 		/// <param name="type">this</param>
 		public static bool IsTypeDef(this Type type) =>
 			!(type is null) && !type.HasElementType && (!type.IsGenericType || type.IsGenericTypeDefinition);
+
+		internal static string Unescape(string name) {
+			if (string.IsNullOrEmpty(name))
+				return name;
+			return name.Replace(@"\,", ",").Replace(@"\+", "+").Replace(@"\&", "&").Replace(@"\*", "*").Replace(@"\[", "[").Replace(@"\]", "]").Replace(@"\\", @"\");
```
0xd4d, line 112: This code is inefficient. You should have a check to see if it has a `\` and if not, return the input string. If there's a `\`, use a string builder.

src/DotNet/ReflectionExtensions.cs
```diff
@@ -105,5 +106,33 @@ static class ReflectionExtensions {
 		/// <param name="type">this</param>
 		public static bool IsTypeDef(this Type type) =>
 			!(type is null) && !type.HasElementType && (!type.IsGenericType || type.IsGenericTypeDefinition);
+
+		internal static string Unescape(string name) {
+			if (string.IsNullOrEmpty(name) || !name.Contains(@"\"))
```
0xd4d, line 111: `name.IndexOf('\\') < 0` is better since it just searches for one char instead of a substring.

CreateAndInject:
> Are method/field/prop/event names also escaped?

No, Reflection doesn't escape these names.

0xd4d:
Thanks, merged!

PR #385, opened 2020-08-14
'=' is not an special char for TypeName

yyjdelete:
Align with the logic in `FullNameFactory.AddIdentifier(string)` and `ReflectionExtensions.IsReservedTypeNameChar(char)` in dnlib and `ResourceRenamer.IsReservedTypeNameChar(char)` in de4dot

### DIFF
--- a/src/DotNet/TypeNameParser.cs
+++ b/src/DotNet/TypeNameParser.cs
@@ -841,7 +841,6 @@ namespace dnlib.DotNet {
 			case '*':
 			case '[':
 			case ']':
-			case '=':
 				return -1;
 
 			default:


0xd4d:
It can't be removed because it's used to stop when there's an `=` in `key=value`, eg. `PublicKeyToken=xxxxx`.

yyjdelete:
Ok, I didn't see the code here.
https://github.com/0xd4d/dnlib/blob/2cbe552792769895991b5e1e3bd130526a12e2ed/src/DotNet/TypeNameParser.cs#L743-L748

But seems the usage below should ignore '='(hit this with renamer in ConfuserEx), maybe needed add another check option or reuse `ignoreWhiteSpace`.
https://github.com/0xd4d/dnlib/blob/2cbe552792769895991b5e1e3bd130526a12e2ed/src/DotNet/TypeNameParser.cs#L358-L363

0xd4d:
What does the type ref name look like and is it a valid name?

If so we could add a new bool arg to `ReadId(bool)`

yyjdelete:
As an example, when you compile the below code and rename `Test` to `Te=st` with dnSpy, you can still execute it without exception and get `ConsoleApp1.Te=st`, but unable to see `[Custom(Type = typeof(Test))]` (only `[Custom]` is shown) in dnSpy.
```cs
using System;
using System.Reflection;

namespace ConsoleApp1
{
	public class Test { }
	[Custom(Type = typeof(Test))]
	public class CustomAttribute : Attribute {
		public Type Type { get; set; }
	}
	class Program {
			Console.WriteLine(typeof(CustomAttribute).GetCustomAttribute<CustomAttribute>().Type);
		}
	}
}
```

0xd4d:
Thanks for the repro. Do you want to update your PR? I think all we need is an extra bool parameter that decides if `=` should be treated as an id char or not.

PR #415, opened 2021-07-12
ResolutionScope now actually can be null/0 in .NET 5 reference assemblies

echalone:
Guess what I found out:
ResolutionScope now actually can be null/0 in .NET 5 reference assemblies, so the solution is to not throw an error and just return 0.
If you want to see this at work just build the attached project and try to rewrite the reference assembly (the one in the /ref subfolder) with this module, it will fail with error message "dnlib.DotNet.Writer.ModuleWriterException: ResolutionScope is null" though everything would be correct if we just return 0 (which is, after all, also what was read in).
I don't know if other properties of MetaData could be affected too, but that's what we experienced so far.
[TestProject.zip](https://github.com/0xd4d/dnlib/files/6799054/TestProject.zip)
Thanks so much for unarchiving the repository for this Pull Request! :)

### DIFF
--- a/src/DotNet/Writer/Metadata.cs
+++ b/src/DotNet/Writer/Metadata.cs
@@ -2325,7 +2325,6 @@ namespace dnlib.DotNet.Writer {
 		/// <returns>Its encoded token</returns>
 		protected uint AddResolutionScope(IResolutionScope rs) {
 			if (rs is null) {
-				// ResolutionScope can now actually be null in .NET 5 Reference Assemblies, so do not throw error any longer here, just return 0
 				return 0;
 			}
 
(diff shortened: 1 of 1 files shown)


wtfsck:
Are you sure this isn't a bug in the C# compiler?

wtfsck:
Seems like it's allowed to be null according to the spec.

> The *TypeRef* table has the following columns:
> *ResolutionScope* (an index into a *Module*, *ModuleRef*, *AssemblyRef* or *TypeRef* table, or null; more precisely, a *ResolutionScope* (§24.2.6) coded index)

src/DotNet/Writer/Metadata.cs
```diff
@@ -2325,7 +2325,7 @@ struct MethodScopeDebugInfo {
 		/// <returns>Its encoded token</returns>
 		protected uint AddResolutionScope(IResolutionScope rs) {
 			if (rs is null) {
-				Error("ResolutionScope is null");
+				// ResolutionScope can now actually be null in .NET 5 Reference Assemblies, so do not throw error any longer here, just return 0
```
wtfsck, line 2328: Since this is allowed according to the spec, this comment can now be removed.

echalone: Thanks for the feedback, I removed the comment again

PR #416, opened 2021-07-29
Preserve Log2Rid value when rewriting a module.

ElektroKill:
Also allows for the user to specify the value.

Can be used to bypass an anti-tampering check in which the obfuscator writes a different value than `1` and then checks at runtime if it is what it wrote.


src/DotNet/Writer/ModuleWriterBase.cs
```diff
@@ -384,6 +384,10 @@ public class ModuleWriterOptionsBase {
 				Cor20HeaderOptions.MinorRuntimeVersion = 5;
 			}
 
+			if (module is ModuleDefMD moduleDefMd) {
```
wtfsck, line 387: There's code below that checks if it's a `ModuleDefMD` so you can move the if block body there instead.

wtfsck (changes requested):
Other than that, LGTM!

wtfsck (approved):

PR #418, opened 2021-08-02
CA and CDI support for manifest resources

ElektroKill:
Add support for reading and writing Custom Attributes and Custom Debug Infos for manifest resources.

### DIFF
--- a/src/DotNet/ModuleDefMD.cs
+++ b/src/DotNet/ModuleDefMD.cs
@@ -1478,17 +1478,17 @@ namespace dnlib.DotNet {
 
 			if (token.Rid == 0) {
 				if (TryCreateResourceStream(mr.Offset, out var dataReaderFactory, out uint resourceOffset, out uint resourceLength))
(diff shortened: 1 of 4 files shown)


ElektroKill:
I've implemented reading in the best way I could think of that matches the implementation for other members and doesn't introduce any API changes except for the fact that `EmbeddedResource`, `AssemblyLinkedResource`, and `LinkedResource` are no longer `sealed` due to inheritance being required by the `MD` variants of them. If there is any way to prevent the non-internal inheritance of a class in this particular case I would be willing to hear about it and implement it.

wtfsck:
> If there is any way to prevent the non-internal inheritance of a class in this particular case I would be willing to hear about it and implement it.

I don't think it's possible so nothing to do about it.

Have you tested this with some assembly that has CAs and/or CDIs?

wtfsck (approved):

ElektroKill:
Yes I have indeed tested this by loading an assembly with resources with dnlib, adding a custom attribute to one of the resources, writing to disk, loading this newly written assembly and checking if the custom attribute is present.

PR #419, opened 2021-08-03
Add support for reading V1 resources

ElektroKill:
Version 1 resources should now be read properly. They are rewritten as version 2 when `ResourceWriter` is used. This code was based on the code in the `ResourceReader` class in `mscorlib` for .NET 2.0. I'm unaware of a way to produce these version 1 resources so I was not able to test this code but since it is pretty much 1:1 to the official implementation I think it's ok.

### DIFF
--- a/src/DotNet/Resources/ResourceReader.cs
+++ b/src/DotNet/Resources/ResourceReader.cs
@@ -201,7 +201,8 @@ namespace dnlib.DotNet.Resources {
 			if (typeIndex < 0 || typeIndex >= userTypes.Count)
 				throw new ResourceReaderException($"Invalid resource type index: {typeIndex}");
 			var type = userTypes[typeIndex];
-			switch (type.Name) {
+			var actualName = type.Name.Split(',')[0];
+			switch (actualName) {
 			case "System.String":   return resourceDataFactory.Create(reader.ReadSerializedString());
 			case "System.Int32":    return resourceDataFactory.Create(reader.ReadInt32());
 			case "System.Byte":     return resourceDataFactory.Create(reader.ReadByte());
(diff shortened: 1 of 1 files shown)


wtfsck:
I'd feel better if this new code has been tested. Not sure where to find old .NET 1 assemblies with these resources though.

src/DotNet/Resources/ResourceReader.cs
```diff
@@ -201,7 +201,8 @@ sealed class ResourceInfo {
 			if (typeIndex < 0 || typeIndex >= userTypes.Count)
 				throw new ResourceReaderException($"Invalid resource type index: {typeIndex}");
 			var type = userTypes[typeIndex];
-			switch (type.Name) {
+			var actualName = type.Name.Split(',')[0];
```
KvanTTT, line 204: It's better to use `count` if possible:

```suggestion
			var actualName = type.Name.Split(',', 2)[0];
```

ElektroKill: That overload doesn't exist :/ 
image
The code would have to be:
`var actualName = type.Name.Split(new[] { ',' }, 2)[0];`
There is also the option of:
`string actualName = type.Name.Substring(0, type.Name.IndexOf(','));`

KvanTTT: Ok then (it seems like my method is available in later versions). The second option is not correct because `,` may be not presented. The following code is correct:

```cs
var commaIndex = type.Name.IndexOf(',');
string actualName = commaIndex == -1 ? type.Name : type.Name.Remove(commaIndex);
```

wtfsck (approved):

wtfsck:
Thanks! It's been merged.

PR #421, opened 2021-12-01
Show method name and token

CreateAndInject:
Show method name and token in order to know which method occurs the error.
I add `args` for `IWriterError`, so de4dot can ignore congener messages

### DIFF
--- a/src/DotNet/Writer/IWriterError.cs
+++ b/src/DotNet/Writer/IWriterError.cs
@@ -12,7 +12,13 @@ namespace dnlib.DotNet.Writer {
 		/// </summary>
 		/// <param name="message">Error message</param>
 		void Error(string message);
+	}
 
+	/// <summary>
+	/// Gets notified of errors. The default handler should normally throw since the written data
+	/// will probably be invalid. Any error can be ignored.
+	/// </summary>
+	public interface IWriterError2 : IWriterError {
 		/// <summary>
 		/// Called when an error is detected (eg. a null pointer or other invalid value). The error
 		/// can be ignored but the written data won't be valid.
--- a/src/DotNet/Writer/IWriterError.cs
+++ b/src/DotNet/Writer/IWriterError.cs
@@ -11,6 +11,7 @@ namespace dnlib.DotNet.Writer {
 		/// can be ignored but the written data won't be valid.
 		/// </summary>
 		/// <param name="message">Error message</param>
-		void Error(string message);
+		/// <param name="args">Optional message arguments</param>
+		void Error(string message, params object[] args);
 	}
 }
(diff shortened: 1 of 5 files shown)


src/DotNet/Writer/IWriterError.cs
```diff
@@ -11,6 +11,7 @@ public interface IWriterError {
 		/// can be ignored but the written data won't be valid.
 		/// </summary>
 		/// <param name="message">Error message</param>
-		void Error(string message);
+		/// <param name="args">Optional message arguments</param>
+		void Error(string message, params object[] args);
```
wtfsck, line 15: This is a breaking change.

wtfsck (changes requested):

src/DotNet/Writer/IWriterError.cs
CreateAndInject: It's compatible on source level, do you mean breaking change on binary?

src/DotNet/Writer/IWriterError.cs
```diff
@@ -12,5 +12,13 @@ public interface IWriterError {
...
 		/// <param name="message">Error message</param>
 		void Error(string message);
+
+		/// <summary>
+		/// Called when an error is detected (eg. a null pointer or other invalid value). The error
+		/// can be ignored but the written data won't be valid.
+		/// </summary>
+		/// <param name="message">Error message</param>
+		/// <param name="args">Optional message arguments</param>
+		void Error(string message, params object[] args);
```
wtfsck, line 22: This is still a breaking change.

You can create a new `IWriteError2` interface and check for it at runtime.

wtfsck (changes requested):

src/DotNet/Writer/MethodBodyWriter.cs
```diff
@@ -8,7 +8,7 @@ namespace dnlib.DotNet.Writer {
 	/// <summary>
 	/// Returns tokens of token types, strings and signatures
 	/// </summary>
-	public interface ITokenProvider : IWriterError {
+	public interface ITokenProvider : IWriterError2 {
```
wtfsck, line 11: This is a breaking change. You have to check for `IWriterError2` at runtime

src/DotNet/Writer/MethodBodyWriter.cs
```diff
@@ -84,6 +107,11 @@ public MethodBodyWriter(ITokenProvider helper, CilBody cilBody, bool keepMaxStac
 			this.helper = helper;
 		}
 
+		internal void Reset(MethodDef method, bool keepMaxStack) {
+			this.method = method;
+			Reset(method.Body, keepMaxStack);
+		}
+
 		internal void Reset(CilBody cilBody, bool keepMaxStack) {
```
wtfsck, line 115: Is this method even called other than from the new `Reset()` method? If it only has one caller, it can be inlined in it.

wtfsck (changes requested):

wtfsck (approved):

wtfsck:
Thanks, merged it!

Issue #423, opened 2021-12-05
Lack checking whether result of ToFileOffset is really valid in file layout image cause ReadCilBody crash


wwh1004:
Crash may happen in this method.
``` cs
public CilBody ReadCilBody(IList<Parameter> parameters, RVA rva, GenericParamContext gpContext) {
	if (rva == 0)
		return new CilBody();

	// Create a full stream so position will be the real position in the file. This
	// is important when reading exception handlers since those must be 4-byte aligned.
	// If we create a partial stream starting from rva, then position will be 0 and always
	// 4-byte aligned. All fat method bodies should be 4-byte aligned, but the CLR doesn't
	// seem to verify it. We must parse the method exactly the way the CLR parses it.
	var reader = metadata.PEImage.CreateReader();
	reader.Position = (uint)metadata.PEImage.ToFileOffset(rva);
	return MethodBodyReader.CreateCilBody(this, reader, parameters, gpContext, Context);
}
```
e.g.
An obfuscated assembly's .enc section: VirtualSize=0x345 , RawSize=0x200
A method body in this assembly: RVA=0x210
PEImage.ToFileOffset regards this method body rva is valid and return converted file offset. In fact, this rva is valid but the file offset is invalid because .enc section is decrypted at runtime and the section size in memory is bigger than in disk. ReadCilBody doesn't notice that and it will cause crash.

**The actual situation is similar as above but more complex.
Currently the critical problem is that we don't know rva or fileoffset is correctly converted by the return value.**

I want to improve rva/fileoffset conversion to fix it.
change
``` cs
/// <summary>
/// Converts an <see cref="RVA"/> to a <see cref="FileOffset"/>
/// </summary>
/// <param name="rva">The RVA to convert</param>
/// <returns>The file offset</returns>
public FileOffset ToFileOffset(RVA rva) {
	var section = ToImageSectionHeader(rva);
	if (section is not null)
		return (FileOffset)(rva - section.VirtualAddress + section.PointerToRawData);
	return (FileOffset)rva;
}
```
to
``` cs
/// <summary>
/// Converts an <see cref="RVA"/> to a <see cref="FileOffset"/>, returns 0 if out of range
/// </summary>
/// <param name="rva">The RVA to convert</param>
/// <returns>The file offset</returns>
public FileOffset ToFileOffset(RVA rva) {
	var section = ToImageSectionHeader(rva);
	if (section is not null) {
[...]

PR #428, opened 2021-12-06
Improve rva/fileoffset conversion and ReadCilBody

wwh1004:
Linked issue: https://github.com/0xd4d/dnlib/issues/423
I have tested these changes by placing new dnlib.dll into dnSpy bin folder

### DIFF
--- a/src/PE/PEInfo.cs
+++ b/src/PE/PEInfo.cs
@@ -5,7 +5,7 @@ using dnlib.IO;
 
 namespace dnlib.PE {
 	/// <summary>
-	/// Reads all PE sections from a PE stream
+	/// Reads all PE sections from a PE stream, for more information see https://docs.microsoft.com/en-us/windows/win32/debug/pe-format
 	/// </summary>
 	sealed class PEInfo {
 		readonly ImageDosHeader imageDosHeader;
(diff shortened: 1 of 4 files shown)


src/PE/PEInfo.cs
```diff
@@ -77,35 +77,54 @@ sealed class PEInfo {
...
 		/// <summary>
-		/// Converts a <see cref="FileOffset"/> to an <see cref="RVA"/>
+		/// Converts a <see cref="FileOffset"/> to an <see cref="RVA"/>, returns 0 if out of range
 		/// </summary>
 		/// <param name="offset">The file offset to convert</param>
 		/// <returns>The RVA</returns>
 		public RVA ToRVA(FileOffset offset) {
+			if (imageSectionHeaders.Length == 0)
+				return (RVA)offset;
+			// In pe headers
```
wtfsck, line 96: It seems like your comments is about the previous line, could you add the comments before the statement, eg.
```suggestion
			if (imageSectionHeaders.Length == 0) {
				// In pe headers
				return (RVA)offset;
			}
```

I also added `{ }` since it now has two lines instead of one.

or 

```C#

			if ((uint)rva >= imageNTHeaders.OptionalHeader.SizeOfImage)
				return 0;
			// Check if rva is larger than memory layout size
->
			// Check if rva is larger than memory layout size
			if ((uint)rva >= imageNTHeaders.OptionalHeader.SizeOfImage)
				return 0;
```

Same thing about the other comments below.

wwh1004: Done.

wtfsck:
Thanks!

PR #429, opened 2021-12-06
Improve Importer TryToUseMethodDefs and TryToUseFieldDefs options

wwh1004:
Obfuscator may rename many methods/fields into same name then TryResolveMethod/TryResolveField will return inconsistent method/field.

### DIFF
--- a/src/DotNet/Importer.cs
+++ b/src/DotNet/Importer.cs
@@ -452,6 +452,13 @@ namespace dnlib.DotNet {
 			if (methodBase is null)
 				return null;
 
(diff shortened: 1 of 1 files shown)


wtfsck:
What's the field or method attributes? Eg. is the method `privatescope` (low 3 bits clear) and is the field also `privatescope` (low 3 bits clear)?

wtfsck (approved):

wwh1004:
I just port my code into dnlib. I don't test it. I will modify it then send or.

PR #437, opened 2021-12-25
Update AssemblyRef.Attributes when updating AssemblyRef.PublicKeyOrToken

### LINKED ISSUE #436: Wrong assembly reference
```
class Program {
		var module = ModuleDefMD.Load(typeof(Program).Module);
		var method = module.EntryPoint.DeclaringType.FindMethod("Test");
		foreach (var instr in method.Body.Instructions) {
			if (instr.OpCode == OpCodes.Ldtoken && instr.Operand is TypeRef tr) {
				string typeFullName = tr.AssemblyQualifiedName; // System.AppDomain, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089
				var type = TypeNameParser.ParseReflection(module, typeFullName, null);
				instr.Operand = type;
				break;
			}
		}
		module.Write("1.exe");
	}

	static void Test() {
		Console.WriteLine(typeof(AppDomain));
		Console.ReadKey();
	}
}
```

The PublicKeyToken of mscorlib should be `b77a5c561934e089`, but dnlib produces `900e13d46cb0307b`:
 
image

CreateAndInject:
Fix #436

### DIFF
--- a/src/DotNet/AssemblyRef.cs
+++ b/src/DotNet/AssemblyRef.cs
@@ -77,7 +77,10 @@ namespace dnlib.DotNet {
 		/// <exception cref="ArgumentNullException">If <paramref name="value"/> is <c>null</c></exception>
 		public PublicKeyBase PublicKeyOrToken {
 			get => publicKeyOrToken;
-			set => publicKeyOrToken = value ?? throw new ArgumentNullException(nameof(value));
+			set {
+				publicKeyOrToken = value ?? throw new ArgumentNullException(nameof(value));
+				ModifyAttributes(value is PublicKey, AssemblyAttributes.PublicKey);
+			}
 		}
 		/// <summary/>
 		protected PublicKeyBase publicKeyOrToken;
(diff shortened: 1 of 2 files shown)


src/DotNet/AssemblyRef.cs
```diff
@@ -77,7 +77,10 @@ public abstract class AssemblyRef : IHasCustomAttribute, IImplementation, IResol
 		/// <exception cref="ArgumentNullException">If <paramref name="value"/> is <c>null</c></exception>
 		public PublicKeyBase PublicKeyOrToken {
 			get => publicKeyOrToken;
-			set => publicKeyOrToken = value ?? throw new ArgumentNullException(nameof(value));
+			set {
+				publicKeyOrToken = value ?? throw new ArgumentNullException(nameof(value));
+				ModifyAttributes(value is PublicKey, AssemblyAttributes.PublicKey);
```
wtfsck, line 82: This property should not touch the assembly attributes.

wtfsck (changes requested):
The correct fix is to fix any code that depend on that flag being set/cleared.

wtfsck (approved):

PR #439, opened 2021-12-28
Fixes calling ResolveTypeSpec without gpContext causing listTypeSpecMD polluted

wwh1004:


wtfsck:
That check was added due to: https://github.com/0xd4d/dnlib/issues/331

Have you tested your code with the code from that issue?

wtfsck:
Seems like the only problem is that it doesn't find the generic parameter in the type so it can't print the GP name, only its GP index. It shouldn't be a problem.

wwh1004:
The TypeSpec I printed is in instruction read by MethodBodyReader, which is read with gpContext. 
https://github.com/0xd4d/dnlib/blob/master/src/Utils/SimpleLazyList.cs#L84
When I resolve TypeSpec without gpContext, 'elem.ContainsGenericParameter' will return false and result will be cached in 'elements[index]'. After that no matter what gpContext I pass to resolve TypeSpec, it will return the cached item but not construct a new one by reading with gpContext.

wwh1004:
> That check was added due to: #331
> 
> Have you tested your code with the code from that issue?

I patched dnSpy's dnlib with this fix and load 'HelloWorld.MaliciousTypeSpecLoop.exe', dnSpy doesn't hang.

wtfsck (approved):

wtfsck:
Thanks for testing it! Merged.

PR #443, opened 2021-12-29
Improve Importer TryToUseMethodDefs and TryToUseFieldDefs options 2

wwh1004:
Obfuscator may rename many methods/fields into same name then TryResolveMethod/TryResolveField will return inconsistent method/field.

Previous PR contains some problems and I fixed it. In previous PR you mentioned 'privatescope' visibility. I check some documents and find it is visible in the same module and can't be referenced in MemberRef table. I think these code won't get error when encountering 'privatescope' visibility. Should it be handled specially?

### DIFF
--- a/src/DotNet/Importer.cs
+++ b/src/DotNet/Importer.cs
@@ -452,6 +452,14 @@ namespace dnlib.DotNet {
 			if (methodBase is null)
 				return null;
 
+			if (TryToUseMethodDefs && IsThisModule(methodBase.Module) &&
+				!methodBase.IsGenericMethod && (methodBase.DeclaringType is null || !methodBase.DeclaringType.IsGenericType) &&
+				module.ResolveToken(methodBase.MetadataToken) is MethodDef md) {
+				// In same module and method and declaring type are both non-generic, directly resolve method definition.
+				// Obfuscator may rename many methods into same name then TryResolveMethod will return inconsistent method.
+				return md;
+			}
+
 			if (forceFixSignature) {
 				//TODO:
 			}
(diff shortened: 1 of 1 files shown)


CreateAndInject:
It's better use `AssemblyNameComparer` or `SigComparer` to determine if is the same Module/Assembly

src/DotNet/Importer.cs
```diff
@@ -508,6 +516,8 @@ public Importer(ModuleDef module, ImporterOptions options, GenericParamContext g
 			}
 		}
 
+		bool IsThisModule(Module module2) => module.Name == module2.Name && IsThisAssembly(module2);
```
wtfsck, line 519: `SigComparer` uses `module2.ScopeName` and it should be case insensitive comparison, see `bool Equals(IModule a, Module b) {`.

wtfsck (changes requested):

wtfsck:
> Previous PR contains some problems and I fixed it. In previous PR you mentioned 'privatescope' visibility. I check some documents and find it is visible in the same module and can't be referenced in MemberRef table. I think these code won't get error when encountering 'privatescope' visibility. Should it be handled specially?

I don't think it's necessary to check for `privatescope` here.

wtfsck:
> It's better to use `AssemblyNameComparer` or `SigComparer` to determine if is the same Module/Assembly

True, but in this case only the name is compared and there's no public API that only compares the name.

src/DotNet/Importer.cs
wwh1004: Done.

wtfsck (approved):

wtfsck:
Thanks, merged!

Issue #427, opened 2021-12-05
Infinite loop in SigComparer.GetHashCode


CreateAndInject:
```C#
class Program {
		static void Main(string[] args) {
			var module = ModuleDefMD.Load(typeof(Program).Module);
			var method = module.EntryPoint.DeclaringType.FindMethod("Test");
			foreach (var instr in method.Body.Instructions) {
				if (instr.Operand is MethodSpec ms) {
					int code = new SigComparer().GetHashCode(ms); // Infinite loop
					Console.WriteLine(code);
				}
			}
			Console.WriteLine("OK");
			Console.ReadKey();
		}

		static void Test<T>(Dictionary<T, T> dict) => string.Concat(dict);
	}
```

wtfsck:
Where in `SigComparer` does it keep looping?

wwh1004:
generic argument of methodspec still contains generic parameter, dnlib will repeat to replace T with KeyValuePair<TKey,TValue> so there is a loop.
https://github.com/0xd4d/dnlib/blob/master/src/DotNet/SigComparer.cs#L2087

CreateAndInject:
> generic argument of methodspec still contains generic parameter, dnlib will repeat to replace T with KeyValuePair<TKey,TValue> so there is a loop. https://github.com/0xd4d/dnlib/blob/master/src/DotNet/SigComparer.cs#L2087

Seems it's not the key point, since it's no problem if modify `Test` like this:
```
static void Test<T>(List<T> list) => string.Concat(list);
```

PR #444, opened 2021-12-30
Fix infinite loop in SigComparer.GetHashCode(TypeSig)

wwh1004:
1.
Fix https://github.com/0xd4d/dnlib/issues/427
In code:
``` cs
static void Test<T>(Dictionary<T, T> dict) { string.Concat(dict); }
```
MethodSig of 'string.Concat' is 'string (IEnumerable<KeyPairValue<T, T>>)' and dnlib will repeat substitute one of 'T' in 'KeyPairValue<T, T>' with ''KeyPairValue<T, T>'' itself, so there is an infinite loop. I add a parameter to disable substituting when get hash code for 'GenericInstSig.GenericArguments'.

2.
About deleted code:
``` cs
if (SubstituteGenericParameters) {
	InitializeGenericArguments();
	genericArguments.PushTypeArgs(gia.GenericArguments);
	hash += GetHashCode(gia.GenericType);
	genericArguments.PopTypeArgs();
}
```
GenericType is always TypeDef or TypeRef so it makes no sense to substitute generic parameters.

### DIFF
--- a/src/DotNet/SigComparer.cs
+++ b/src/DotNet/SigComparer.cs
@@ -2007,8 +2007,11 @@ exit: ;
 				return 0;
 			int hash;
 
-			if (substituteGenericParameters && genericArguments is not null)
+			if (substituteGenericParameters && genericArguments is not null) {
+				var t = a;
 				a = genericArguments.Resolve(a);
+				substituteGenericParameters = t == a;
+			}
 
 			switch (a.ElementType) {
 			case ElementType.Void:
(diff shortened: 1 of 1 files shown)


CreateAndInject:
image
Nice, but can you ensure this?

wtfsck:
Thanks for the PR, will check later when I have more time.

wwh1004:
Still contains some problems. WIP

wwh1004:
Currently it will not resolve 'T' which is already resolved. In reflection apis, ga is always closed instantiation. So we don't need to substitute an already resolved ga.

wwh1004:
In theory, these changes will not affect SigComparer's handling of signatures which can be fully represented by System.Reflection types, since System.Reflection types do not allow open instantiation. It should ensure that the original code behavior is not changed. Namely it just fixes this issue without side effects.

wtfsck:
Merged now, thanks!

Issue #238, opened 2019-01-29
Show more information.


CreateAndInject:
The most frequent error of dnlib is:
Error calculating max stack value....

Now, dnSpy & de4dot ref dnlib by nuget, not source, everytime I must debug de4dot.exe in dnSpy and breaking in WriteMethodBodies to view the method name. It's so troublesome.

1. Can you restore dnlib source instead of nuget? Not everyone is so skilled as you, we must debug with dnlib source to resolve our problem.

2.Can you show method name when occurs: Error calculating max stack value....

0xd4d:
1: no

2: can do that.

riQQ:
Shouldn't https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg solve 1?

0xd4d:
The nuget package already has sourcelink support so it should work today, yes. The pdb file is included with the nupkg. Next release will use snupkg files.

MeikTranel:
Check @shanselman blog for both producing and consuming these symbols:

https://www.hanselman.com/blog/ExploringNETCoresSourceLinkSteppingIntoTheSourceCodeOfNuGetPackagesYouDontOwn.aspx

@CreateAndInject, try enabling the Visual Studio options (the changes to the CSProj are only if you want to produce packages with sourcelink enabled)

CreateAndInject:
@MeikTranel How?

MeikTranel:
It is explained in the article i liked. I also made a Video showing how to do it with DNLib (Ignore the error frame... that wouldnt happen to you):

https://gfycat.com/marriedimpressionablecaecilian

CreateAndInject:
@MeikTranel Thanks

CreateAndInject:
Any news about `Can you show method name when occurs: Error calculating max stack value....` ?

By the way:
1. Please update sourcelink, or else dnlib can't build:
https://github.com/dotnet/sourcelink/issues/337#issue-466979138
https://github.com/aspnet/EntityFrameworkCore/issues/16656#issue-469885486
https://github.com/dotnet/sourcelink/issues/350#issue-470675436

2. I notice you make a large changes in dnlib & dnSpy about `Pattern Matching(use is null instead of ==)` & `Nullable reference type`, etc.
Please don't do it in de4dot, or else there will be large conflicts when modded version pull it.

0xd4d:
Don't hijack old issues with new issues. Create new issues instead.

[...]

PR #447, opened 2021-12-30
Improve the error output that occurs when saving the module

wwh1004:
Fixes https://github.com/0xd4d/dnlib/issues/238

Test code:
``` cs
using System;
using System.IO;
using dnlib.DotNet;
using dnlib.DotNet.Emit;
using dnlib.DotNet.Writer;

class Program {
	static void Main() {
		Test1();
		Console.ReadKey(true);
	}

	static void Test1() {
		using var module = ModuleDefMD.Load(typeof(Program).Module);
		var type = module.FindNormal("Program");
		var method = type.FindMethod("Test1");

		module.Resources.Add(null);
		// null resource

		type.CustomAttributes.Add(null);
		// null custom attribute

		method.ParamDefs.Add(new ParamDefUser { Constant = new ConstantUser() });
		// null constant

		var body = method.Body;
		for (int i = 0; i < ushort.MaxValue + 10; i++)
			body.Variables.Add(new Local(module.CorLibTypes.Int32));
		// too many locals

		body.Instructions[0] = new Instruction(OpCodes.Ldstr, 123);
		// mismatching operand type

		body.Instructions[1] = new Instruction(OpCodes.Pop);
		body.Instructions[2] = new Instruction(OpCodes.Pop);
		// evaluation stack overflow

		Test2();
		type.Remove(type.FindMethod("Test2"));
		// removed method but still referenced

		using var output = new MemoryStream();
		var writerOptions = new ModuleWriterOptions(module) { Logger = new Logger() };
		module.Write(output, writerOptions);
	}

	static void Test2() {
	}

	class Logger : ILogger {
		int count = 0;

		public bool IgnoresEvent(LoggerEvent loggerEvent) {
			return loggerEvent != LoggerEvent.Error && loggerEvent != LoggerEvent.Warning;
		}

		public void Log(object sender, LoggerEvent loggerEvent, string format, params object[] args) {
			Console.WriteLine($"{++count}. {string.Format(format, args)}");
			Console.WriteLine();
		}
	}
}
```

Before:
```
1. Constant is null

2. Custom attribute is null

3. Resource is null

4. Error calculating max stack value. If the method's obfuscated, set CilBody.KeepOldMaxStack or MetadataOptions.Flags (KeepOldMaxStack, global option) to ignore this error. Otherwise fix your generated CIL code so it conforms to the ECMA standard. At method System.Void Program::Test1() (06000002).

5. Too many locals, max number of locals is 65535 (0xFFFF)

6. Invalid instruction operand

7. Method System.Void Test2() (06000003) is not defined in this module (dnlib-test.dll). A method was removed that is still referenced by this module.
```

After:
```
1. Constant is null. Error occurred after metadata event MemberDefRidsAllocated during writing method 'System.Void Program::Test1()' (0x06000002).

2. Custom attribute is null. Error occurred after metadata event MostTablesSorted during writing type 'Program' (0x02000002).

3. Resource is null. Error occurred after metadata event BeginAddResources.

4. Error calculating max stack value. If the method's obfuscated, set CilBody.KeepOldMaxStack or MetadataOptions.Flags (KeepOldMaxStack, global option) to ignore this error. Otherwise fix your generated CIL code so it conforms to the ECMA standard. Error occurred after metadata event BeginWriteMethodBodies during writing method 'System.Void Program::Test1()' (0x06000002).

5. Too many locals, max number of locals is 65535 (0xFFFF). Error occurred after metadata event BeginWriteMethodBodies during writing method 'System.Void Program::Test1()' (0x06000002).

6. Invalid instruction operand. Error occurred after metadata event BeginWriteMethodBodies during writing method 'System.Void Program::Test1()' (0x06000002).

7. Method 'System.Void Test2()' (0x06000003) is not defined in this module 'dnlib-test.dll'. A method was removed that is still referenced by this module. Error occurred after metadata event BeginWriteMethodBodies during writing method 'System.Void Program::Test1()' (0x06000002).
```

If allows break changes in new major version, is it possible to merge IWriterError2 into IWriterError and delete Extensions.Error2? I think code can be more clear in that case.
https://github.com/wwh1004/dnlib/blob/a9bb8f9e7514504e22b9d98d2ac625d5b4d975bb/src/DotNet/Writer/IWriterError.cs#L39

### DIFF
--- a/src/DotNet/Writer/IWriterError.cs
+++ b/src/DotNet/Writer/IWriterError.cs
@@ -36,7 +36,7 @@ namespace dnlib.DotNet.Writer {
 		/// <param name="helper">The instance of <see cref="IWriterError"/></param>
 		/// <param name="message">Error message</param>
 		/// <param name="args">Optional message arguments</param>
-		public static void Error2(this IWriterError helper, string message, params object[] args) {
+		internal static void Error2(this IWriterError helper, string message, params object[] args) {
 			if (helper is IWriterError2 helper2)
 				helper2.Error(message, args);
 			else
--- a/src/DotNet/Writer/IWriterError.cs
+++ b/src/DotNet/Writer/IWriterError.cs
@@ -27,4 +27,20 @@ namespace dnlib.DotNet.Writer {
 		/// <param name="args">Optional message arguments</param>
 		void Error(string message, params object[] args);
 	}
+
+	static partial class Extensions {
+		/// <summary>
+		/// Called when an error is detected (eg. a null pointer or other invalid value). The error
+		/// can be ignored but the written data won't be valid.
+		/// </summary>
+		/// <param name="helper">The instance of <see cref="IWriterError"/></param>
+		/// <param name="message">Error message</param>
+		/// <param name="args">Optional message arguments</param>
+		public static void Error2(this IWriterError helper, string message, params object[] args) {
+			if (helper is IWriterError2 helper2)
+				helper2.Error(message, args);
+			else
+				helper.Error(string.Format(message, args));
+		}
+	}
 }
--- a/src/DotNet/Writer/Metadata.cs
+++ b/src/DotNet/Writer/Metadata.cs
@@ -1582,7 +1582,7 @@ namespace dnlib.DotNet.Writer {
 			int notifyAfter = numTypes / numNotifyEvents;
 
 			foreach (var type in allTypeDefs) {
-				errorContext.Source = type;
+				using var _ = errorContext.SetSource(type);
 
 				if (typeNum++ == notifyAfter && notifyNum < numNotifyEvents) {
 					RaiseProgress(Writer.MetadataEvent.MemberDefRidsAllocated, (double)typeNum / numTypes);
@@ -1612,7 +1612,7 @@ namespace dnlib.DotNet.Writer {
 						Error("Field is null");
 						continue;
 					}
-					errorContext.Source = field;
+					using var __ = errorContext.SetSource(field);
 					uint rid = GetRid(field);
 					var row = new RawFieldRow((ushort)field.Attributes, stringsHeap.Add(field.Name), GetSignature(field.Signature));
 					tablesHeap.FieldTable[rid] = row;
@@ -1621,7 +1621,6 @@ namespace dnlib.DotNet.Writer {
 					AddFieldRVA(field);
 					AddImplMap(new MDToken(Table.Field, rid), field);
 					AddConstant(new MDToken(Table.Field, rid), field);
-					errorContext.Source = null;
 				}
 
 				var methods = type.Methods;
@@ -1632,7 +1631,7 @@ namespace dnlib.DotNet.Writer {
 						Error("Method is null");
 						continue;
 					}
-					errorContext.Source = method;
+					using var __ = errorContext.SetSource(method);
 					if (method.ExportInfo is not null)
 						ExportedMethods.Add(method);
 					uint rid = GetRid(method);
@@ -1657,7 +1656,6 @@ namespace dnlib.DotNet.Writer {
 						AddConstant(new MDToken(Table.Param, pdRid), pd);
 						AddFieldMarshal(new MDToken(Table.Param, pdRid), pd);
 					}
-					errorContext.Source = null;
 				}
 
 				var events = type.Events;
@@ -1668,12 +1666,11 @@ namespace dnlib.DotNet.Writer {
 						Error("Event is null");
 						continue;
 					}
-					errorContext.Source = evt;
+					using var __ = errorContext.SetSource(evt);
 					uint rid = GetRid(evt);
 					var row = new RawEventRow((ushort)evt.Attributes, stringsHeap.Add(evt.Name), AddTypeDefOrRef(evt.EventType));
 					tablesHeap.EventTable[rid] = row;
 					AddMethodSemantics(evt);
-					errorContext.Source = null;
 				}
 
 				var properties = type.Properties;
@@ -1684,16 +1681,13 @@ namespace dnlib.DotNet.Writer {
 						Error("Property is null");
 						continue;
 					}
-					errorContext.Source = prop;
+					using var __ = errorContext.SetSource(prop);
 					uint rid = GetRid(prop);
 					var row = new RawPropertyRow((ushort)prop.Attributes, stringsHeap.Add(prop.Name), GetSignature(prop.Type));
 					tablesHeap.PropertyTable[rid] = row;
 					AddConstant(new MDToken(Table.Property, rid), prop);
 					AddMethodSemantics(prop);
-					errorContext.Source = null;
 				}
-
-				errorContext.Source = null;
 			}
 		}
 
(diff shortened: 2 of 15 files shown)


wtfsck:
I'll check this later when I have more time. I just had a very quick check.

- I think this fixes https://github.com/0xd4d/dnlib/issues/238 , and if so add `Fixes https://github.com/0xd4d/dnlib/issues/238` to your first post.
- I think the responsibility of adding the `.` is in the other code, not the callers of `Error()`. There will be less changes and we won't forget to add a period in the future if we add another error.

src/DotNet/Writer/IWriterError.cs
```diff
@@ -27,4 +27,20 @@ public interface IWriterError2 : IWriterError {
...
+
+	static partial class Extensions {
+		/// <summary>
+		/// Called when an error is detected (eg. a null pointer or other invalid value). The error
+		/// can be ignored but the written data won't be valid.
+		/// </summary>
+		/// <param name="helper">The instance of <see cref="IWriterError"/></param>
+		/// <param name="message">Error message</param>
+		/// <param name="args">Optional message arguments</param>
+		public static void Error2(this IWriterError helper, string message, params object[] args) {
```
wtfsck, line 39: Does this need to be public?

src/DotNet/Writer/Metadata.cs
```diff
@@ -445,6 +445,7 @@ public abstract class Metadata : IReuseChunk, ISignatureWriterHelper, ITokenProv
 		RVA rva;
 		readonly MetadataOptions options;
 		ILogger logger;
+		internal readonly MetadataErrorContext errorContext;
```
wtfsck, line 448: Does this need to be `internal`?

src/DotNet/Writer/MetadataErrorContext.cs
```diff
@@ -0,0 +1,46 @@
+using System;
```
wtfsck, line 1: Need the license header here, see the other files.

src/DotNet/Writer/MetadataErrorContext.cs
```diff
@@ -0,0 +1,46 @@
...
+			newArgs[ctxArgIndex] = Event;
+
+			if (tokenSource is not null) {
+				string sourceType = tokenSource switch {
+					TypeDef => "type",
+					FieldDef => "field",
+					MethodDef => "method",
+					EventDef => "event",
+					PropertyDef => "property",
+					_ => throw new InvalidOperationException()
```
wtfsck, line 32: We should not throw here since if this happens, it's a bug in our code, not the users and it's just here to make a better exception message. Instead use something else such as `"???"`.

src/DotNet/Writer/MetadataErrorContext.cs
```diff
@@ -0,0 +1,46 @@
...
+				};
+				newMessage.AppendFormat(" during writing {0} '{{{1}}}' (0x{{{2}:X8}})", sourceType, ctxArgIndex + 1, ctxArgIndex + 2);
+				newArgs[ctxArgIndex + 1] = tokenSource;
+				newArgs[ctxArgIndex + 2] = tokenSource.MDToken.Raw;
+			}
+			else if (stringSource is not null) {
+				newMessage.AppendFormat(" during writing {0}", stringSource);
```
wtfsck, line 39: What happens if `stringSource` already has `{0}` etc? It probably needs to be escaped.

src/DotNet/Writer/MethodBodyWriter.cs
```diff
@@ -58,29 +57,6 @@ public sealed class MethodBodyWriter : MethodBodyWriterBase {
 		/// </summary>
 		public uint LocalVarSigTok => localVarSigTok;
 
-		/// <summary>
-		/// Constructor
-		/// </summary>
-		/// <param name="helper">Helps this instance</param>
-		/// <param name="method">The method</param>
-		public MethodBodyWriter(ITokenProvider helper, MethodDef method)
```
wtfsck, line 66: These two public constructors can't be removed, since it's a breaking change. Unless we wait to merge this PR so it's added in v4.0.

Deprecating them is an option. They're relatively new so should have few users.

src/DotNet/Writer/PreserveTokensMetadata.cs
```diff
@@ -855,7 +855,7 @@ public PreserveTokensMetadata(ModuleDef module, UniqueChunkList<ByteArrayChunk>
 			foreach (var type in allTypeDefs) {
 				if (type is null)
 					continue;
-
+
```
wtfsck, line 858: Useless whitespace.

wtfsck (commented):
Looks good, just a few problems, including the extra `"."`.

src/DotNet/Writer/MetadataErrorContext.cs
wwh1004: 'stringSource' shouldn't appear '{?}', variable 'stringSource' is not error message but just respresents error source. e.g.: "exported types", "entry point", "vtable fixups".

src/DotNet/Writer/Metadata.cs
wwh1004: MetadataErrorContext is internal type and I don't know whether child type like NormalMetadata will use errorContext to mark error source in the future. So I set errorContext to internal.

src/DotNet/Writer/MethodBodyWriter.cs
wwh1004: If wait v4, do we need merge IWriterError2 into IWriterError and delete Extensions.Error2 as breaking changes?

src/DotNet/Writer/Metadata.cs
wtfsck: `internal` can be added if needed in the future if that ever happens. It's better to not add `internal` if not needed.

src/DotNet/Writer/MetadataErrorContext.cs
wtfsck: This wasn't fixed so I've 'unresolved' this.

src/DotNet/Writer/MethodBodyWriter.cs
wtfsck: We don't need to wait for v4 just keep both interface, but when v4 is released, these interfaces can be merged.

wtfsck:
See also my comments here: https://github.com/0xd4d/dnlib/pull/447#issuecomment-1003106625

CreateAndInject:
@wtfsck Can you complete this PR yourself or bump version without this PR? This PR is sended more than 2 months, maybe @wwh1004 stoped working on it.

wtfsck:
@wwh1004 are you still working on this PR?

wwh1004:
> @wwh1004 are you still working on this PR?
I will update it these days.

wwh1004:
Updated

src/DotNet/Writer/Metadata.cs
```diff
@@ -1608,16 +1621,18 @@ internal sealed class Rows<T> where T : class {
 					AddFieldRVA(field);
 					AddImplMap(new MDToken(Table.Field, rid), field);
 					AddConstant(new MDToken(Table.Field, rid), field);
+					errorContext.Source = null;
```
wtfsck, line 1624: This code should initialize `Source` to the previous value which was initialized above to`errorContext.Source = type;`. There's other code in this file that also overwrites it with null when the old value should be used.

The best solution is to create a new disposable `struct` that initializes `Source` and saves the original value and restores it in `Dispose()`.

wwh1004:
Updated

wtfsck:
Thanks!

PR #448, opened 2022-01-05
Validate data in case allocate too many memory, fix #442

### LINKED ISSUE #442: dnlib x64 exhausts all memory when read method body.
[test.zip](https://github.com/0xd4d/dnlib/files/7789825/test.zip)
```
        Console.WriteLine(IntPtr.Size); // Can only reproduce if current program is AnyCPU/x64
	var module = ModuleDefMD.Load("test.exe");
	var method = module.ResolveMethod(0x11d);
	var body = method.Body; // here
        Console.WriteLine(body);
	Console.ReadKey();
}
```

CreateAndInject:
Fix #442

### DIFF
--- a/src/DotNet/CustomAttributeReader.cs
+++ b/src/DotNet/CustomAttributeReader.cs
@@ -277,7 +277,7 @@ namespace dnlib.DotNet {
 		}
 
 		List<CANamedArgument> ReadNamedArguments(int numNamedArgs) {
-			if ((uint)numNamedArgs > 0x4000_0000 || numNamedArgs * 4 > reader.BytesLeft)
+			if ((uint)numNamedArgs >= 0x4000_0000 || numNamedArgs * 4 > reader.BytesLeft)
 				return null;
 			var namedArgs = new List<CANamedArgument>(numNamedArgs);
 			for (int i = 0; i < numNamedArgs; i++) {
--- a/src/DotNet/CustomAttributeReader.cs
+++ b/src/DotNet/CustomAttributeReader.cs
@@ -208,8 +208,6 @@ namespace dnlib.DotNet {
 		/// <returns>A list of <see cref="CANamedArgument"/>s or <c>null</c> if some error
 		/// occurred.</returns>
 		internal static List<CANamedArgument> ReadNamedArguments(ModuleDef module, ref DataReader reader, int numNamedArgs, GenericParamContext gpContext) {
-			if (numNamedArgs > 0x10000)
-				return null;
 			try {
 				var caReader = new CustomAttributeReader(module, ref reader, gpContext);
 				var namedArgs = caReader.ReadNamedArguments(numNamedArgs);
@@ -279,6 +277,8 @@ namespace dnlib.DotNet {
 		}
 
 		List<CANamedArgument> ReadNamedArguments(int numNamedArgs) {
+			if ((uint)numNamedArgs > 0x4000_0000 || numNamedArgs * 4 > reader.BytesLeft)
+				return null;
 			var namedArgs = new List<CANamedArgument>(numNamedArgs);
 			for (int i = 0; i < numNamedArgs; i++) {
 				if (reader.Position == reader.Length)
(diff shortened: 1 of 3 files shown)


src/DotNet/CustomAttributeReader.cs
```diff
@@ -208,6 +208,8 @@ public struct CustomAttributeReader {
 		/// <returns>A list of <see cref="CANamedArgument"/>s or <c>null</c> if some error
 		/// occurred.</returns>
 		internal static List<CANamedArgument> ReadNamedArguments(ModuleDef module, ref DataReader reader, int numNamedArgs, GenericParamContext gpContext) {
+			if (numNamedArgs > 0x10000)
```
wtfsck, line 211: The check should be in `ReadNamedArguments()`.

It should not limit it to 64K args. You should instead check that there's enough bytes left to read all named args. Each named arg is at least 4 bytes in size so make sure count is not negative and also < 0x4000_0000 and also that there's at least count*4 bytes left in the reader: `(uint)count < 0x4000_0000 && count * 4 < bytes left`

src/DotNet/CustomAttributeReader.cs
```diff
@@ -528,7 +530,7 @@ public struct CustomAttributeReader {
 			int arrayCount = reader.ReadInt32();
 			if (arrayCount == -1) {	// -1 if it's null
 			}
-			else if (arrayCount < 0)
+			else if (arrayCount < 0 || arrayCount > 0x10000)
```
wtfsck, line 533: This should compare it against 'bytes left' and not limit it to 64K

src/DotNet/SignatureReader.cs
```diff
@@ -482,12 +482,12 @@ public struct SignatureReader {
...
 		T ReadSig<T>(T methodSig) where T : MethodBaseSig {
 			if (methodSig.Generic) {
-				if (!reader.TryReadCompressedUInt32(out uint count))
+				if (!reader.TryReadCompressedUInt32(out uint count) || count > 0x10000)
 					return null;
 				methodSig.GenParamCount = count;
 			}
 
-			if (!reader.TryReadCompressedUInt32(out uint numParams))
+			if (!reader.TryReadCompressedUInt32(out uint numParams) || numParams > 0x10000)
```
wtfsck, line 490: Need this: `count > bytes left` (if true, fail).

src/DotNet/SignatureReader.cs
```diff
@@ -513,7 +513,7 @@ public struct SignatureReader {
 		/// <param name="callingConvention">First byte of signature</param>
 		/// <returns>A new <see cref="LocalSig"/> instance</returns>
 		LocalSig ReadLocalSig(CallingConvention callingConvention) {
-			if (!reader.TryReadCompressedUInt32(out uint count))
+			if (!reader.TryReadCompressedUInt32(out uint count) || count > 0x10000)
```
wtfsck, line 516: Need this: `count > bytes left` (if true, fail).

src/DotNet/SignatureReader.cs
```diff
@@ -528,7 +528,7 @@ public struct SignatureReader {
 		/// <param name="callingConvention">First byte of signature</param>
 		/// <returns>A new <see cref="GenericInstMethodSig"/> instance</returns>
 		GenericInstMethodSig ReadGenericInstMethod(CallingConvention callingConvention) {
-			if (!reader.TryReadCompressedUInt32(out uint count))
+			if (!reader.TryReadCompressedUInt32(out uint count) || count > 0x10000)
```
wtfsck, line 531: Need this: `count > bytes left` (if true, fail).

src/DotNet/SignatureReader.cs
```diff
@@ -605,7 +605,7 @@ public struct SignatureReader {
 
 			case ElementType.GenericInst:
 				nextType = ReadType();
-				if (!reader.TryReadCompressedUInt32(out num))
+				if (!reader.TryReadCompressedUInt32(out num) || num > 0x10000)
```
wtfsck, line 608: Need this: `count > bytes left` (if true, fail).

wtfsck (changes requested):

src/DotNet/CustomAttributeReader.cs
```diff
@@ -277,6 +277,8 @@ public struct CustomAttributeReader {
 		}
 
 		List<CANamedArgument> ReadNamedArguments(int numNamedArgs) {
+			if ((uint)numNamedArgs > 0x4000_0000 || numNamedArgs * 4 > reader.BytesLeft)
```
wtfsck, line 280: `=>`: `numNamedArgs >= 0x4000_0000`

wtfsck (changes requested):

wtfsck (approved):

wtfsck:
Thanks, merged!

PR #449, opened 2022-01-07
Use RuntimeInformation class to check environment.

ElektroKill:


wtfsck (approved):

wtfsck:
LGTM. Does this fix a problem you've discovered?

ElektroKill:
I’m using windows so this change does not affect me. The reason for this change is that with .NET Standard, dnlib can run on Unix systems under .NET Core and not Mono so I added the proper .NET Standard way of checking the platform.

wtfsck:
Thanks!

PR #470, opened 2022-06-03
Improvements to `DynamicMethodBodyReader`

ElektroKill:
* Added support for reading `InitLocals` property and use the value in `GetMethod` and `RestoreMethod`.
* Set `MaxStack` property in `RestoreMethod`.

### DIFF
--- a/src/DotNet/Emit/DynamicMethodBodyReader.cs
+++ b/src/DotNet/Emit/DynamicMethodBodyReader.cs
@@ -62,6 +62,7 @@ namespace dnlib.DotNet.Emit {
 		readonly MethodDef method;
 		readonly int codeSize;
 		readonly int maxStack;
+		readonly bool? initLocals;
 		readonly List<object> tokens;
 		readonly IList<object> ehInfos;
 		readonly byte[] ehHeader;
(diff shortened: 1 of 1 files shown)


src/DotNet/Emit/DynamicMethodBodyReader.cs
```diff
@@ -534,6 +537,16 @@ class ExceptionInfo {
...
 		}
 
+		/// <inheritdoc />
+		public override void RestoreMethod(MethodDef method) {
+			base.RestoreMethod(method);
+
+			var body = method.Body;
+			if (initLocals.HasValue)
+				body.InitLocals = initLocals.Value;
+			body.MaxStack = (ushort)Math.Min(maxStack, ushort.MaxValue);
```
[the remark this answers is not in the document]
ElektroKill: I added the code to initialize MaxStack so that when the using the preserve maxstack metadata writer option is used the max stack is still a proper value. The reason why for the cast is that I copied the line from https://github.com/0xd4d/dnlib/pull/470/files#diff-5a6271fc9c1b45930fcd380798359844d3431c2c3f0b48a56fa8325f44975edcR395

src/DotNet/Emit/DynamicMethodBodyReader.cs
```diff
@@ -186,9 +187,10 @@ public DynamicMethodBodyReader(ModuleDef module, object obj, Importer importer,
 			if (code is null)
 				throw new Exception("No code");
 			codeSize = code.Length;
-			var delMethod = rslvMethodFieldInfo.Read(obj) as SR.MethodBase;
+			var delMethod = rslvMethodFieldInfo.Read(obj) as DynamicMethod;
```
wtfsck, line 190: Did you test .NET Framework 3.5 and 4.x?

ElektroKill: I checked the source code of `mscorlib` using dnSpy for both 2.0.0.0 and 4.0.0.0 versions and the field is of type `DynamicMethod`.

wtfsck (approved):

wtfsck:
Thanks, merged!

PR #473, opened 2022-07-28
fix: correct a wrong flag in IsClass

gdh1995:
The old usage of `TypeAttributes.Interface` in `public bool IsClass` is a typo bug.


CreateAndInject:
1. The original code isn't wrong
2. `attributes & TypeAttributes.Class` is meaningless, since `TypeAttributes.Class` is `0`, so the result is always `0`
3. Test first when you send PR

gdh1995:
sorry...

artemking4:
Just stumbled upon that. Indeed, the original code might not be wrong, but the description is misleading then. It might need a change.
image

PR #474, opened 2022-08-13
Fix ArgumentOutOfRangeException

CreateAndInject:


wtfsck:
Looks like a caller bug not a dnlib bug.

CreateAndInject:
> Looks like a caller bug not a dnlib bug.

Almost all extension methods in the dnlib never occur extentions
For example : 
```csharp
public static TypeSig GetNext(this TypeSig self) => self?.Next;
```
The only purpose of this method is : The called is unnecessary to check null when calling

What's why I prevent possible exception to stay the same action to other extension methods

PR #503, opened 2023-06-10
Support new `PrimaryConstructorInformationBlob` Roslyn CDI

ElektroKill:


wtfsck:
Does this blob contain any metadata tokens?

ElektroKill:
> Does this blob contain any metadata tokens?

Glancing at Roslyn source code, it seams to me that this blob is, at least for now, always empty.
Code which adds the CDI in Roslyn:
https://github.com/dotnet/roslyn/blob/5100eeedfe43c45b53fc298d6341d4d52e0d735a/src/Compilers/Core/Portable/PEWriter/MetadataWriter.PortablePdb.cs#L144-L150
Code which reads it (according to GitHub search there is only one place where it is read and that is the Debugger EE)
https://github.com/dotnet/roslyn/blob/5100eeedfe43c45b53fc298d6341d4d52e0d735a/src/ExpressionEvaluator/Core/Source/ExpressionCompiler/PDB/MethodDebugInfo.Portable.cs#L401

wtfsck (approved):

PR #504, opened 2023-06-10
Add rudimentary support for `DeserializingResourceReader`

ElektroKill:
* Updated code in `CheckReaders` to match the actual meaning of the fields. The implementation now matches the one in the runtime.
* Added several new members to allow creating of more customized resource sets.
* Added support for the `DeserializingResourceReader` class introduced in .NET 5 and now in the `System.Resources.Extensions` package.

The changes should be source and binary compatible with the previous dnlib version. This limited the support I could provide for this new resource reader which specifies additional information on how to deserialize objects with different serialization kinds for the `UserType` type code.

See:
https://github.com/dotnet/dotnet/blob/95703b57cf3dfaa79f054d277551a809d4f7696a/src/runtime/src/libraries/System.Resources.Extensions/src/System/Resources/Extensions/DeserializingResourceReader.cs#L108

The old `ResourceReader` always used `BinaryFormater` while the new code will first read some additional information and then deserialize the object. This new implementation in dnlib leaves the handling of this to the user which could be confusing. Ideally, we'd want to handle that in dnlib itself but that can't be done without introducing breaking changes. 

Would it be OK for us to introduce breaking changes as part of a minor release or would this have to be part of a major release? I'd like to use this new code in [dnSpyEx](https://github.com/dnSpyEx) and would like to know if I should wait for the next minor release or perhaps, implement a custom version of the `ResourceReader` resource handling in dnSpyEx temporarily. If breaking changes are OK in this component, I'd also like to extend the writing support to support version 1 resources!

### DIFF
--- a/src/DotNet/Resources/ResourceBinaryWriter.cs
+++ b/src/DotNet/Resources/ResourceBinaryWriter.cs
@@ -1,0 +1,27 @@
+using System.IO;
+
+namespace dnlib.DotNet.Resources {
+	/// <summary>
+	/// Extension of <see cref="BinaryWriter"/> for writing resource set elements
+	/// </summary>
+	public sealed class ResourceBinaryWriter : BinaryWriter {
+		/// <summary>
+		/// Format version of the resource set
+		/// </summary>
+		public int FormatVersion { get; internal set; }
+
+		/// <summary>
+		/// Specifies the target reader type of the resource set
+		/// </summary>
+		public ResourceReaderType ReaderType { get; internal set; }
+
+		/// <inheritdoc />
+		public ResourceBinaryWriter(Stream stream) : base(stream) { }
+
+		/// <summary>
+		/// Writes a 7-bit encoded integer.
+		/// </summary>
+		/// <param name="value">The value to write</param>
+		public new void Write7BitEncodedInt(int value) => base.Write7BitEncodedInt(value);
+	}
+}
--- a/src/DotNet/Resources/ResourceBinaryWriter.cs
+++ b/src/DotNet/Resources/ResourceBinaryWriter.cs
@@ -15,8 +15,7 @@ namespace dnlib.DotNet.Resources {
 		/// </summary>
 		public ResourceReaderType ReaderType { get; internal set; }
 
-		/// <inheritdoc />
-		public ResourceBinaryWriter(Stream stream) : base(stream) { }
+		internal ResourceBinaryWriter(Stream stream) : base(stream) { }
 
 		/// <summary>
 		/// Writes a 7-bit encoded integer.
--- a/src/DotNet/Resources/ResourceDataFactory.cs
+++ b/src/DotNet/Resources/ResourceDataFactory.cs
@@ -173,18 +173,18 @@ namespace dnlib.DotNet.Resources {
 		/// <param name="value">Serialized data</param>
 		/// <param name="type">Type of serialized data</param>
 		/// <returns></returns>
-		public BinaryResourceData CreateSerialized(byte[] value, UserResourceType type) => new BinaryResourceData(CreateUserResourceType(type.Name, true), value);
+		public BinaryResourceData CreateSerialized(byte[] value, SerializationFormat format, UserResourceType type) => new BinaryResourceData(CreateUserResourceType(type.Name, true), value, format);
 
 		/// <summary>
 		/// Creates serialized data
 		/// </summary>
 		/// <param name="value">Serialized data</param>
 		/// <returns></returns>
-		public BinaryResourceData CreateSerialized(byte[] value) {
+		public BinaryResourceData CreateBinaryFormatterSerialized(byte[] value) {
 			if (!GetSerializedTypeAndAssemblyName(value, out var assemblyName, out var typeName))
 				throw new ApplicationException("Could not get serialized type name");
 			string fullName = $"{typeName}, {assemblyName}";
-			return new BinaryResourceData(CreateUserResourceType(fullName), value);
+			return new BinaryResourceData(CreateUserResourceType(fullName), value, SerializationFormat.BinaryFormatter);
 		}
 
 		sealed class MyBinder : SerializationBinder {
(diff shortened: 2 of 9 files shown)


src/DotNet/Resources/ResourceElementSet.cs
```diff
@@ -10,6 +11,37 @@ namespace dnlib.DotNet.Resources {
...
 
+		/// <summary>
+		/// The ResourceReader type name used by this set
+		/// </summary>
+		public string ResourceReaderTypeName { get; internal set; }
+
+		/// <summary>
+		/// Gets whether <see cref="ResourceReaderTypeName"/> is a DeserializingResourceReader
+		/// </summary>
+		public bool UsesDeserializingResourceReader => ResourceReaderTypeName is not null && Regex.IsMatch(ResourceReaderTypeName, @"^System\.Resources\.Extensions\.DeserializingResourceReader,\s*System\.Resources\.Extensions");
```
wtfsck, line 22: Do we need to check for a comma or something at the end of the string?

wtfsck:
> Would it be OK for us to introduce breaking changes as part of a minor release or would this have to be part of a major release?

It needs to be a new major release. Do you have any other breaking changes you'd like to see?

ElektroKill:
> > Would it be OK for us to introduce breaking changes as part of a minor release or would this have to be part of a major release?
> 
> It needs to be a new major release. Do you have any other breaking changes you'd like to see?

See https://github.com/0xd4d/dnlib/issues/509, other than that no potential breaking changes come to mind as of now.

Should I proceed with implementing the breaking changes necessary for the proper support of this new resource reader class in this PR?

wtfsck:
> Should I proceed with implementing the breaking changes necessary for the proper support of this new resource reader class in this PR?

Yes, next release will bump the major version number.

ElektroKill:
The two latest commits add proper support for this new resource reader type. Something which is still TBD is writing format version 1 resources.

src/DotNet/Resources/ResourceBinaryWriter.cs
```diff
@@ -0,0 +1,27 @@
...
+		/// </summary>
+		public int FormatVersion { get; internal set; }
+
+		/// <summary>
+		/// Specifies the target reader type of the resource set
+		/// </summary>
+		public ResourceReaderType ReaderType { get; internal set; }
+
+		/// <inheritdoc />
+		public ResourceBinaryWriter(Stream stream) : base(stream) { }
```
wtfsck, line 19: This ctor is public but the properties above only has internal setters. Seems like the ctor should be internal.

src/DotNet/Resources/ResourceDataFactory.cs
```diff
@@ -173,18 +173,18 @@ public class ResourceDataFactory {
 		/// <param name="value">Serialized data</param>
 		/// <param name="type">Type of serialized data</param>
 		/// <returns></returns>
-		public BinaryResourceData CreateSerialized(byte[] value, UserResourceType type) => new BinaryResourceData(CreateUserResourceType(type.Name, true), value);
+		public BinaryResourceData CreateSerialized(byte[] value, SerializationFormat format, UserResourceType type) => new BinaryResourceData(CreateUserResourceType(type.Name, true), value, format);
```
wtfsck, line 176: XML doc is missing `format` arg.

src/DotNet/Resources/ResourceReader.cs
```diff
@@ -45,7 +46,7 @@ public ResourceReaderException(SerializationInfo info, StreamingContext context)
 	/// <param name="type">Serialized type</param>
 	/// <param name="serializedData">Serialized data</param>
 	/// <returns></returns>
-	public delegate IResourceData CreateResourceDataDelegate(ResourceDataFactory resourceDataFactory, UserResourceType type, byte[] serializedData);
+	public delegate IResourceData CreateResourceDataDelegate(ResourceDataFactory resourceDataFactory, UserResourceType type, byte[] serializedData, SerializationFormat format);
```
wtfsck, line 49: XML doc is missing `format` arg.

src/DotNet/Resources/ResourceWriter.cs
```diff
@@ -78,20 +83,12 @@ public sealed class ResourceWriter {
 			writer.Write(dataStream.ToArray());
 		}
 
-		void WriteData(BinaryWriter writer, ResourceElement info, IFormatter formatter) {
+		void WriteData(ResourceBinaryWriter writer, ResourceElement info, IFormatter formatter) {
 			var code = GetResourceType(info.ResourceData);
-			WriteUInt32(writer, (uint)code);
+			writer.Write7BitEncodedInt((int)code);
```
wtfsck, line 88: Is this identical to the original code?

src/DotNet/Resources/ResourceReader.cs
```diff
@@ -219,52 +220,51 @@ sealed class ResourceInfo {
...
+		IResourceData ReadSerializedObject(uint endPos, ResourceReaderType readerType, UserResourceType type) {
+			byte[] serializedData;
+			IResourceData res;
+			switch (readerType) {
+			case ResourceReaderType.ResourceReader:
+				serializedData = reader.ReadBytes((int)(endPos - reader.Position));
+				res = createResourceDataDelegate?.Invoke(resourceDataFactory, type, serializedData, SerializationFormat.BinaryFormatter);
+				return res ?? resourceDataFactory.CreateSerialized(serializedData, SerializationFormat.BinaryFormatter, type);
+			case ResourceReaderType.DeserializingResourceReader: {
+				var format = (SerializationFormat)reader.Read7BitEncodedInt32();
```
wtfsck, line 236: This enum value is read from user data so we should verify that it's a valid value before passing it to the following functions.

src/DotNet/Resources/ResourceWriter.cs
ElektroKill: Yes, the new method is implemented as follows in the core library:
```csharp
protected void Write7BitEncodedInt(int value)
{
	uint num;
	for (num = (uint) value; num >= 128U; num >>= 7)
		this.Write((byte) (num | 128U));
	this.Write((byte) num);
}
```

This new method is called with the enum casted to `int` is equivalent to the previous implementation which casts the enum to `uint` because the core library method casts the `int` to a `uint` itself.
```csharp
static void WriteUInt32(BinaryWriter writer, uint value) {
	while (value >= 0x80) {
		writer.Write((byte)(value | 0x80));
		value >>= 7;
	}
	writer.Write((byte)value);
}
```

wtfsck (approved):

Issue #514, opened 2023-07-12
Consider adding a net6.0 as an additional target framework


ElektroKill:
The new versions of .NET offer additional API's which can be used by Roslyn when compiling C# code to generate more performant code. One of the most notable changes for codegen is the use of string-builder-like on stack structs for string interpolation instead of the classic concatenation calls or format calls.

Adding `net6.0` as a TFM would allow Roslyn to apply these optimizations when compiling for that target. `dnlib` consumers could then benefit from improved string performance when using dnlib.

Note: I have no ran any benchmarks and the performance improvements are theoretical.

ElektroKill:
If this is green-lighted, I can contribute this change.

wtfsck:
Sure. I just checked the csproj file and we have this, does it also need to be updated?

```csproj
  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
    <PackageReference Include="System.Reflection.Emit" Version="4.7.0" />
    <PackageReference Include="System.Reflection.Emit.Lightweight" Version="4.7.0" />
  </ItemGroup>
```

PR #515, opened 2023-07-15
Add additional net6.0 target framework

ElektroKill:
Fixes https://github.com/0xd4d/dnlib/issues/514

This PR adds support `net6.0` target for `dnlib`. The rationale for this change was discussed in the aforementioned issue and updates the NuGet packages used by `dnlib` to the latest available versions.

I've also gone ahead and replaced most of the usages of string concatenation and replaced it with string interpolation. This allows Roslyn to better take advantage of the faster string API's available in .NET 6. Some code changes were also necessary to resolve warnings regarding obsolete members and platform compatibility of the unmanaged Windows PDB support. A few NRE's which were picked up by code analysis were also fixed!

### DIFF
--- a/src/DotNet/ModuleDefMD.cs
+++ b/src/DotNet/ModuleDefMD.cs
@@ -1495,7 +1495,12 @@ namespace dnlib.DotNet {
 			return new EmbeddedResourceMD(this, mr, Array2.Empty<byte>());
 		}
 
-		[HandleProcessCorruptedStateExceptions, SecurityCritical]	// Req'd on .NET Framework 4.0
+		// Required attributes on .NET Framework 4.0
+		// HandleProcessCorruptedStateExceptions is obsolete on .NET Core and newer
+#if !NETCOREAPP
+		[HandleProcessCorruptedStateExceptions]
+#endif
+		[SecurityCritical]
 		bool TryCreateResourceStream(uint offset, out DataReaderFactory dataReaderFactory, out uint resourceOffset, out uint resourceLength) {
 			dataReaderFactory = null;
 			resourceOffset = 0;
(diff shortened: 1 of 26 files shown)


src/DotNet/Pdb/Managed/DbiModule.cs
```diff
@@ -157,7 +157,7 @@ sealed class DbiModule {
 			if (funcs[found].Lines is null) {
 				while (found > 0) {
 					var prevFunc = funcs[found - 1];
-					if (prevFunc is not null || prevFunc.Address != address)
+					if (prevFunc is not null && prevFunc.Address != address)
```
ElektroKill, line 160: I'm pretty sure this was an unintentional mixup of the `||` and `&&` operators which would result in right-hand side being executed only if `prevFunc` was `null` causing an NRE. If I'm wrong feel free to correct me.

src/DotNet/ModuleDefMD.cs
```diff
@@ -1495,7 +1495,12 @@ public sealed class ModuleDefMD : ModuleDefMD2, IInstructionOperandResolver {
 			return new EmbeddedResourceMD(this, mr, Array2.Empty<byte>());
 		}
 
-		[HandleProcessCorruptedStateExceptions, SecurityCritical]	// Req'd on .NET Framework 4.0
+		// Required attributes on .NET Framework 4.0
+		// HandleProcessCorruptedStateExceptions is obsolete on .NET Core and newer
+#if !NETCOREAPP
+		[HandleProcessCorruptedStateExceptions]
+#endif
+		[SecurityCritical]
```
ElektroKill, line 1503: I've gone ahead and used conditional compilation here to exclude the `HandleProcessCorruptedStateExceptions` attribute as it is marked with `Obsolete` and the obsolescence message claims this attribute is ignored by the runtime. I have not verified this behavior. An alternative could be to use `#pragma` directive to disable the obsolescence warning and still include the attribute in the compiled code just in case it does actually make a difference.

wtfsck (approved):

wtfsck:
Thanks!

PR #517, opened 2023-07-20
Compare reference only for MemberDefs if they are from the same module

### LINKED ISSUE #450: Why does dnlib compare two MemberDef by name?
Why `a == b` is not enough?
image

```
class Program
{
    int a;
    int b;
}
```
Rename b to a by dnSpy, then SigCompare will think it's the same field, but it's not in fact.

CreateAndInject:
Fix #450

### DIFF
--- a/src/DotNet/SigComparer.cs
+++ b/src/DotNet/SigComparer.cs
@@ -485,6 +485,12 @@ namespace dnlib.DotNet {
 		/// When comparing types, don't compare a multi-dimensional array's lower bounds and sizes
 		/// </summary>
 		IgnoreMultiDimensionalArrayLowerBoundsAndSizes = 0x800000,
+
+		/// <summary>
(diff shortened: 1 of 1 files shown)


ElektroKill:
I suggest we add a additional signature compared flag to revert to the old behavior to allow easy adaptation of code from v3 to v4.

CreateAndInject:
>It's useful when you open the same assembly but two different versions and need to compare defs.

We should add an option if we don't check if they are from the same module to fix above issue.

But if we already checked the module, no reason to add such option, since it just fix the bug but doesn't take any side effect, unless we can imagine an example which old code work fine but new code make wrong.

wtfsck:
Yes an option to revert to the old behavior is needed.

> since it just fix the bug

It's not a bug as explained in the linked issue.

wtfsck (approved):

wtfsck:
Thanks!

PR #520, opened 2023-08-02
Optimize type name computation

### LINKED ISSUE #519: Optimize properties on `TypeRef` and `TypeDef`
`TypeRef` and `TypeDef` both currently implement `IType` and due to this, they have to provide `string Namespace { get; }` and `string TypeName { get; }` properties. These properties currently call `FullNameFactory.Name` or `FullNameFactory.Namespace`. These calls are expensive as they allocate a `StringBuilder` only to access and append a single similarly named property of type `UTF8String` from the `TypeDef` or `TypeRef`. Modifying these properties to just return the direct value of `UTF8String` properties through the use of its implicit cast operator would help improve the performance of the `IType` implementation in these two types.

If a green light is given, I'd be willing to open a pull request with these changes.

EDIT: After further investigation, this optimization could also be applied to `ExportedType`.

ElektroKill:
Implementation as described in #519 

Will add the assembly full name improvements after they are greenlit over at #519.

fixes #519 (For GitHub to properly link the PR to the issue :D)

### DIFF
--- a/src/DotNet/AssemblyDef.cs
+++ b/src/DotNet/AssemblyDef.cs
@@ -503,14 +503,12 @@ namespace dnlib.DotNet {
 		/// <summary>
 		/// Gets the assembly name with the public key
 		/// </summary>
-		public string GetFullNameWithPublicKey() => GetFullName(publicKey);
+		public string GetFullNameWithPublicKey() => FullNameFactory.AssemblyFullName(this, false);
 
 		/// <summary>
 		/// Gets the assembly name with the public key token
 		/// </summary>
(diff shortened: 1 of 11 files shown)


wtfsck:
> Will add the assembly full name improvements after they are greenlit over at https://github.com/0xd4d/dnlib/issues/519.

Yes, you can improve that code!

wtfsck (approved):
LGTM. Have you written assemblies to disk and compared the files before/after these changes (old dnlig vs new one) to make sure the output didn't change?

ElektroKill:
> LGTM. Have you written assemblies to disk and compared the files before/after these changes (old dnlib vs new one) to make sure the output didn't change?

Yes I have

PR #527, opened 2023-09-19
Optimize finding CustomAttribute by a full name

AlanLiu90:
This PR improves `CustomAttributeCollection.Find`, `CustomAttributeCollection.FindAll` and `CustomAttributeCollection.RemoveAll` by comparing the suffix of the full name first. It can avoid calling `CustomAttribute.TypeFullName` in most of the time.

In a workflow I'm using currently, this PR can reduce the total running time by 29%.


wtfsck:
The args (`string fullName`) could be `null` before but will now throw if `null` so we need a `null` check at the beginning of the methods now.

AlanLiu90:
I updated the code to check if `fullName` is null in these methods.

wtfsck:
Thanks, merged!

PR #528, opened 2023-10-01
Speed up Import(FieldInfo) when debugging

CreateAndInject:
I already reported this issue to Microsoft, see [here](https://github.com/dotnet/runtime/issues/92856)

### DIFF
--- a/src/DotNet/Importer.cs
+++ b/src/DotNet/Importer.cs
@@ -650,7 +650,7 @@ namespace dnlib.DotNet {
 			try {
 				// Get the original field def in case the declaring type is a generic
 				// type instance and the field uses a generic type parameter.
-				origField = fieldInfo.Module.ResolveField(fieldInfo.MetadataToken);
+				origField = fieldInfo.DeclaringType is { IsGenericType: true } ? fieldInfo.Module.ResolveField(fieldInfo.MetadataToken) : fieldInfo;
 			}
 			catch (ArgumentException) {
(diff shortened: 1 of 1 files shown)


CreateAndInject:
Seems this issue only happen when **_unchecked_** `Enable Just My code`

src/DotNet/Importer.cs
```diff
@@ -650,7 +650,7 @@ public Importer(ModuleDef module, ImporterOptions options, GenericParamContext g
 			try {
 				// Get the original field def in case the declaring type is a generic
 				// type instance and the field uses a generic type parameter.
-				origField = fieldInfo.Module.ResolveField(fieldInfo.MetadataToken);
+				origField = fieldInfo.DeclaringType is { IsGenericType: true } ? fieldInfo.Module.ResolveField(fieldInfo.MetadataToken) : fieldInfo;
```
wtfsck, line 653: Is this code exactly identical to the original code?

CreateAndInject: I think so, `typeof(C1<>.C2).IsGenericType` is `true`, so if `IsGenericType` is `false`, I don't think `ResolveField` can return a different value
```C#
public class C1<T>
{
    public class C2
    {
        public T Field;
    }
}
```

wtfsck: I can only accept this PR if it's guaranteed to work the same as before.

PR #529, opened 2023-10-11
Improve IsValueType for GenericSig

CreateAndInject:


wtfsck:
Can you explain why this is an improvement?

CreateAndInject:
Since we know if it's ValueType exactly in these cases.

CreateAndInject:
Reflection check type constraint also:
image

wtfsck:
This resolves a type which no other function in that file does. It's better if you create an extension method to your code and add this code to it.

PR #540, opened 2024-01-24
Clean moduleSearchPaths

CreateAndInject:
By the way, is there any issue in .editorconfig file? The code is always changed when formating (Ctrl + E, D)

### DIFF
--- a/src/DotNet/AssemblyResolver.cs
+++ b/src/DotNet/AssemblyResolver.cs
@@ -336,6 +336,8 @@ namespace dnlib.DotNet {
 #if THREAD_SAFE
 			theLock.EnterWriteLock(); try {
 #endif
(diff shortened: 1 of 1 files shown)


wtfsck:
> By the way, is there any issue in .editorconfig file? The code is always changed when formating (Ctrl + E, D)

It's probably the `#if THREAD_SAFE` stuff that the formatter doesn't like.

wtfsck:
Thanks, merged!

CreateAndInject:
Mark: Fix OutOfMemoryException when handle hundreds of files one by one in 32bit process.

CreateAndInject:
@wtfsck Can you bump version? I need this commit to fix OutOfMemoryException.

wtfsck:
@CreateAndInject Done

PR #554, opened 2024-04-21
Update comment for DynamicMethodBodyReaderOptions

### LINKED ISSUE #552: I already know how UnknownDeclaringType happen
It happen when the developer call `DynamicILInfo.GetTokenFor(RuntimeFieldHandle field)` rather than `DynamicILInfo.GetTokenFor(RuntimeFieldHandle field, RuntimeTypeHandle contextType)` to build DynamicMethod, .NET doesn't crash without declaring type, it can work fine also.
Especially, .NET2.0 doesn't support a public overload with RuntimeTypeHandle parameter, that's why so many protectors doesn't pass in this parameter.

As the reason is clear, such name/comment should be updated.

https://github.com/0xd4d/dnlib/blob/0b2dc951e3a74cd31e177fae62cc56dab676583e/src/DotNet/Emit/DynamicMethodBodyReader.cs#L16-L28
https://github.com/0xd4d/dnlib/blob/0b2dc951e3a74cd31e177fae62cc56dab676583e/src/DotNet/Emit/DynamicMethodBodyReader.cs#L458-L462
https://github.com/0xd4d/dnlib/blob/0b2dc951e3a74cd31e177fae62cc56dab676583e/src/DotNet/Emit/DynamicMethodBodyReader.cs#L458-L462

CreateAndInject:
Close #552

### DIFF
--- a/src/DotNet/Emit/DynamicMethodBodyReader.cs
+++ b/src/DotNet/Emit/DynamicMethodBodyReader.cs
@@ -458,22 +458,22 @@ namespace dnlib.DotNet.Emit {
 				if ((options & DynamicMethodBodyReaderOptions.UnknownDeclaringType) != 0) {
 					// Sometimes it's a generic type but obj != `GenericMethodInfo`, so pass in 'default' and the
 					// runtime will try to figure out the declaring type. https://github.com/0xd4d/dnlib/issues/298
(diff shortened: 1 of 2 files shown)


wtfsck:
This PR does more than the PR title says. Why did you add new code?

wtfsck:
I see, seems like it's based on your other PR https://github.com/0xd4d/dnlib/pull/553 .

You should create a new branch with just the comment fix and force push.

PR #555, opened 2024-04-21
Pass in genericParamProvider

CreateAndInject:


src/DotNet/Importer.cs
```diff
@@ -827,11 +833,11 @@ public Importer(ModuleDef module, ImporterOptions options, GenericParamContext g
 			case ElementType.ByRef:		result = new ByRefSig(Import(type.Next)); break;
 			case ElementType.ValueType: result = CreateClassOrValueType((type as ClassOrValueTypeSig).TypeDefOrRef, true); break;
 			case ElementType.Class:		result = CreateClassOrValueType((type as ClassOrValueTypeSig).TypeDefOrRef, false); break;
-			case ElementType.Var:		result = new GenericVar((type as GenericVar).Number, gpContext.Type); break;
+			case ElementType.Var:		result = new GenericVar((type as GenericVar).Number, gpContext.Type ?? (type as GenericVar).OwnerType); break;
```
wtfsck, line 836: `type as GenericVar` could be `null`. Same below (see MVar below).

wtfsck (changes requested):
Explain what this PR improves.

CreateAndInject:
> Could you explain what this PR improves?

Some code depends on `IMethod.FullName`:
```if (method.FullName == "System.Void System.Collections.Generic.ICollection`1::Add(T)")```
FullName will be ```System.Void System.Collections.Generic.ICollection`1::Add(!0)``` without onwer.

File: src/DotNet/Importer.cs (hunk 1 above)
Comment (CreateAndInject) [reply src/DotNet/Importer.cs:836]: But dnlib already uses `(type as GenericVar).Number`, if it's null, NRE already be thrown before my code get executed.

**Comment by wwh1004** [comment]
If uses dnlib to import a TypeSig from one ModuleDefMD to another ModuleDefMD, we shouldn't assume the imported TypeSig is in the same gp context. Even in the same ModuleDefMD, just importing a TypeSig from one type to another, the gp context is indeterminate.

**Comment by CreateAndInject** [comment]
> If uses dnlib to import a TypeSig from one ModuleDefMD to another ModuleDefMD, we shouldn't assume the imported TypeSig is in the same gp context. Even in the same ModuleDefMD, just importing a TypeSig from one type to another, the gp context is indeterminate.

I think GenericParamContext is designed as:
```
new Importer(module, new GenericParamContext(member1context)).Import(member1);
new Importer(module, new GenericParamContext(member2context)).Import(member2);
...
```
rather than use:
```
module.Import(member1);
module.Import(member2);
...
```

In this PR, preparing different Importers/GenericParamContexts are unnecessary in some cases, just module.Import(...)


> This pr will result in not being able to get a TypeSig with an empty gp context (gp context is read-only, GenericSig cannot be modified after it has been created).

I don't understand, can you explain more?

wwh1004:
gp context is a parameter designed for display. You can leave it empty.

wwh1004:
Almost all existing code does not pass gp context, even de4dot, ConfuserEx. When the gp context is empty, it cannot be assumed that the gp context of TypeSig is consistent with the original one.
Suppose someone uses dnlib to import the TypeSig of module A into module B, and then disposes the ModuleDefMD of A. Then there will be a disposed object in the imported TypeSig in module B. So everyone must use the Importer's constructor with gp context overload. This will change everyone's usage habits.

PR #557, opened 2024-05-28
Add new `AllowByRefLike` generic parameter attribute

ElektroKill:
See:
https://github.com/dotnet/runtime/blob/main/docs/design/features/byreflike-generics.md
https://github.com/dotnet/runtime/issues/102671
https://github.com/dotnet/runtime/blob/9daa4b41eb9f157e79eaf05e2f7451c9c8f6dbdc/src/coreclr/inc/corhdr.h#L847

The code responsible for this flag appears to be in the main branch of the runtime already.

### DIFF
--- a/src/DotNet/GenericParam.cs
+++ b/src/DotNet/GenericParam.cs
@@ -260,6 +260,14 @@ namespace dnlib.DotNet {
 			set => ModifyAttributes(value, GenericParamAttributes.DefaultConstructorConstraint);
 		}
 
+		/// <summary>
+		/// Gets/sets the <see cref="GenericParamAttributes.AllowByRefLike"/> bit
+		/// </summary>
+		public bool AllowsByRefLike {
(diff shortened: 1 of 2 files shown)


wtfsck (changes requested):
This mask needs to be updated too, it's now 0x3C according to the header file you linked to.

```C#
SpecialConstraintMask	= 0x001C,
```

BTW, what happened to the other PR, seems like someone deleted it.

ElektroKill:
> This mask needs to be updated too, it's now 0x3C according to the header file you linked to.
> 
> ```cs
> SpecialConstraintMask	= 0x001C,
> ```

Oh yeah, I missed that! The code should be correct now.

> BTW, what happened to the other PR, seems like someone deleted it.

To be honest, I have no clue. The GitHub account associated with it appears to have been deleted in a way that has deleted all content from the account including comments, etc. instead of replacing them with the typical @ghost user.

wtfsck (approved):

wtfsck:
Thanks, merged!

Issue #558, opened 2024-05-28
Sort `ExportedType` table in the same way as the `TypeDef` table


ElektroKill:
Microsoft has modified the specification to state that the table containing `ExportedType` rows should receive the exact same sorting as the `TypeDef` table.

See: https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md#metadata-logical-format

wtfsck:
I think this only requires fixing `AddExportedTypes()` and just updating the code `var exportedTypes = module.ExportedTypes;` to make sure `exportedTypes` is sorted correctly, similar to what `module.GetTypes()` already does, see `NormalMetadata.GetAllTypeDefs()`

ElektroKill:
Yeah, I thought about just modifying the code in `AddExportedTypes` to sort the exported types in the order that is defined in the ECMA augments documents but I haven't though of a nice and optimal algorithm for sorting. We can't do what we do for normal types since `ExportedType` defines its parent type compared to `TypeDef` which defines all of its nested types making creating the necessary ordering easier. If you have any ideas, lmk! Maybe I'm not thinking of something obvious :D

wtfsck:
If it's a nested exported type, `ExportedType.Implementation` is an `ExportedType`. Just like if a type is nested, `TypeDef.DeclaringType` is a `TypeDef`.

It should be possible to build a temporary `TypeDef.NestedTypes` (for exported types) with a dictionary `ExportedType` -> `List<ExportedType>` by just checking if the current exported type (looping over all of them) is a nested type, and if so, just add it to `dict[parent].Add(current exported type)`. Then once we have that dict, we should be able to just use the same algorithm as module.GetTypes(), except that if `ExportedType.Implementation` is a `FileDef` or `AssemblyRef`, just add it to the result (no special ordering required).

```
foreach et in all-exported-types {
  if (et.Implementation is FileDef or AssemblyRef)
    just add it, no ordering required
  else if it's an ExportedType parent
    dict[parent].Add(et)
}
// now use the same algo as module.GetBytes() but use `dict` instead of checking `NestedTypes` property.
```

Haven't tested this but it seems like it could work.

PR #559, opened 2024-05-29
Sort `ExportedType` table when writing to comply with new ECMA Augments

ElektroKill:
fixes https://github.com/0xd4d/dnlib/issues/558

The algorithm suggested in the issue was adjusted to allow more than one nested type per exported type.

Something to consider:
The current algorithm does not dispose of the enumerators generated just like the method in https://github.com/0xd4d/dnlib/blob/4e0837cb0f4319ffbcdd1642d1973fce203b2177/src/DotNet/AllTypesHelper.cs#L14. This should be fine since list enumerators don't contain any logic in their `Dipose` method but this is something we'd probably need to add to both implementations for future-proofing and potential reuse. Perhaps we could also extract this traversal into an additional class for future use.

### DIFF
--- a/src/DotNet/Writer/Metadata.cs
+++ b/src/DotNet/Writer/Metadata.cs
@@ -1811,9 +1811,39 @@ namespace dnlib.DotNet.Writer {
 		void AddExportedTypes() {
 			using var _ = errorContext.SetSource("exported types");
 			var exportedTypes = module.ExportedTypes;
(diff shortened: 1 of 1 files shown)


wtfsck (approved):

wtfsck:
Thanks, added an issue about disposing of enumerators: https://github.com/0xd4d/dnlib/issues/560

PR #564, opened 2024-11-30
`ModuleWriter` no longer ignores `Cor20HeaderOptions.EntryPoint` value

ElektroKill:
The behavior now matches that of `NativeModuleWriter`


ElektroKill:
CI fails because of a deprecated version of `upload-artifact`

wtfsck:
Yeah, if it's an easy fix, want to fix that too?

wtfsck:
Looks like actions/upload-artifact@v4 is the latest version

PR #567, opened 2025-09-03
Fix potential infinite loop in `TryGetOriginalTargetFrameworkAttribute`

ElektroKill:
This PR fixes a potential infinite loop that could occur when a custom `IAssemblyResolver` implementation calls `TryGetOriginalTargetFrameworkAttribute` on the source module. If this method encounters a generic attribute, upon resolving its name, it will call the `MemberRef` constructor, which will call `ResolveTypeDef`, which can create an infinite loop and lead to a stack overflow.

Since the TargetFrameworkAttribute is not generic, a simple and easy fix is to just skip the processing of all attributes that are generic using the same condition as used in the `MemberRef` constructor to prevent the `ResolveTypeDef` call.


wtfsck:
Thanks, merged!

PR #568, opened 2025-10-12
Fix AssemblyResolver.Remove for fuzzy find

CreateAndInject:


wtfsck:
Late, but it's merged now.

PR #571, opened 2026-02-12
Add MethodImplAttributes.Async flag

PhilippNaused:
Add the new Async flag for methods that was introduced for the new [Runtime-Async Feature](https://github.com/dotnet/runtime/issues/109632) in .NET 11.

See: 
<https://github.com/dotnet/runtime/blob/main/docs/design/specs/runtime-async.md>

<https://github.com/dotnet/runtime/blob/0c7ef81aa7d25dcac00eb81d156e03f4c0ca396b/src/coreclr/inc/corhdr.h#L650>

<https://github.com/dotnet/runtime/blob/0c7ef81aa7d25dcac00eb81d156e03f4c0ca396b/src/libraries/System.Private.CoreLib/src/System/Reflection/MethodImplAttributes.cs#L34>

### DIFF
--- a/src/DotNet/MethodDef.cs
+++ b/src/DotNet/MethodDef.cs
@@ -856,6 +856,14 @@ namespace dnlib.DotNet {
 		}
 
 		/// <summary>
+		/// Gets/sets the <see cref="MethodImplAttributes.Async"/> bit
+		/// </summary>
+		public bool IsAsync {
+			get => ((MethodImplAttributes)implAttributes & MethodImplAttributes.Async) != 0;
+			set => ModifyImplAttributes(value, MethodImplAttributes.Async);
+		}
+
+		/// <summary>
 		/// Gets/sets the <see cref="MethodSemanticsAttributes.Setter"/> bit
 		/// </summary>
 		public bool IsSetter {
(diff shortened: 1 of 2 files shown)


ElektroKill:
`MethodDef.HasReturnType` could also potentially be updated to match this statement:

"Async methods also do not have matching return type conventions as sync methods. For sync methods, the stack should contain a value convertible to the stated return type before the `ret` instruction. For async methods, the stack should be empty in the case of `Task` or `ValueTask`, or the type argument in the case of `Task<T>` or `ValueTask<T>`."

found in https://github.com/dotnet/runtime/blob/main/docs/design/specs/runtime-async.md#i845-sync-and-async-methods

@wtfsck What do you think?

PhilippNaused:
@ElektroKill Like this?

```diff
 /// <summary>
 /// <c>true</c> if the method returns a value (i.e., return type is not <see cref="System.Void"/>)
 /// </summary>
+/// <remarks>
+/// <see cref="MethodImplAttributes.Async"/> methods also do not have matching return type conventions as sync methods.
+/// For sync methods, the stack should contain a value convertible to the stated return type before the <see cref="OpCodes.Ret"/> instruction.
+/// For async methods, the stack should be empty in the case of <c>Task</c> or <c>ValueTask</c>, or the type argument in the case of <c>Task&lt;T&gt;</c> or <c>ValueTask&lt;T&gt;</c>.
+/// <seealso href="https://github.com/dotnet/runtime/blob/main/docs/design/specs/runtime-async.md"/>
+/// </remarks>
 public bool HasReturnType => ReturnType.RemovePinnedAndModifiers().GetElementType() != ElementType.Void;
```

wtfsck:
Yeah you cand add it to the PR.

wtfsck:
Merged, thanks!

PR #573, opened 2026-03-18
Add assembly resolution for .NET runtime

kant2002:
Also make sure that System.Runtime do not get redirected.

### DIFF
--- a/dnlib.sln
+++ b/dnlib.sln
@@ -7,23 +7,62 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "dnlib", "src\dnlib.csproj",
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Examples", "Examples\Examples.csproj", "{F27E72B5-C4BD-40BF-AD19-4C8A99B55872}"
 EndProject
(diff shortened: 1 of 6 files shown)


kant2002:
This change blocks my ability to port ConfuserEx to Linux/Mac so I would appreciate that somehow it will trigger faster release.

ElektroKill:
Hello,

This is not a valid way to fix assembly resolution for .NET Core/Standard/.NET 5+

The current assembly resolver is built to follow assembly resolution logic and rules of .NET Framework only. Adding a search path to the current runtime directory does not fix the problem, nor does the redirect. These are just hack fixes.

A proper solution is to determine the runtime of the file that the assembly reference originates from and use the appropriate logic.

The assembly resolution logic for .NET Core and similar is very different, as it does not search the GAC at all and instead searches the path in the .NET installation directory for the appropriate .NET version. Resolving .NET Core 3.0 files to .NET 7 files, for example, is completely wrong. Things like the runtime WindowsDesktop vs. regular vs. AspNetCore need to be considered too.

kant2002:
I don't brave enough to implement all that logic without having test for existing behavior, so I going for smallest fix. I do acknowledge that this is not full fix, but what's action plan. Implement everything at once? How work can be split and delivered in small increments?

kant2002:
Is there any simpler things on path for proper implementation? I willing to contribute but I don't have capacity make it in one go.

PR #574, opened 2026-05-01
Modernize examples

kant2002:
- Use .NET 10, since .NET 6 is not by default on new computer anymore and it's not supported
- You cannot save to drive C: in Windows 11 by default. Maybe just my computer, who knowns. Anyway, that make examples Windows only, which is unnecessary
- Example3 crashed, since invariant changed, and you should pass empty string as locale.

### DIFF
--- a/Examples/Example2.cs
+++ b/Examples/Example2.cs
@@ -56,7 +56,7 @@ namespace dnlib.Examples {
 			body.Instructions.Add(OpCodes.Ret.ToInstruction());
 
 			// Save the assembly to a file on disk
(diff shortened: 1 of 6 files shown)


kant2002:
@ElektroKill can you take a look?

ElektroKill:
> @ElektroKill can you take a look?

Changes look good to me, the rest is dependent on @wtfsck.

Since .NET 6 is out of support updating dnlib to target `netstandard2.0;net45;net10.0` instead of `netstandard2.0;net45;net6.0` would be a good change too I think.

wtfsck:
Thanks, I merged it.

PR #576, opened 2026-05-31
Preserve the type name string in `CustomMarshalType` if parsing it fails

ElektroKill:
Currently, if the type name was not parsed, any empty string was written.

This has the side effect that code relying on `GetCustomAttribute<MarshalAsAttribute>()` will throw a different exception.

For example:
`System.ArgumentException: The name of the type is invalid. (Parameter 'typeName@0')`
vs.
`System.ArgumentException: The value cannot be an empty string. (Parameter 'typeName')`

This behavior could be used to detect rewriting with dnlib as a means of anti-tampering protection.


wtfsck (approved):

wtfsck:
Thanks, merged.

